PR #44, opened 2018-09-18
Remove "restart: always" from docker-compose files

brikou:
"restart: always" would make services start (again) whenever you boot your laptop, I don't think this is the expected behaviour.

### DIFF
--- a/flow-graphql/docker-compose.yml
+++ b/flow-graphql/docker-compose.yml
@@ -2,9 +2,8 @@ version: '3'
 services:
   prisma:
     image: prismagraphql/prisma:1.17-beta-1
-    restart: always
     ports:
-    - "4466:4466"
+      - "4466:4466"
     environment:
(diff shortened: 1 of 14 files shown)


schickling:
@dpetrick what are your thoughts on this?

dpetrick:
We can have the examples use the `on-failure` restart policy, which I _assume_ doesn't start the container on reboot, but finding details about that is as difficult as always in the docker docs.

brikou:
I think that people using docker in production knows about restart option. And if Prisma crashes it is better to not hide it. It is dev afterall.

abhiaiyer91:
+1 for on-failure

marktani:
Why did you close this?

brikou:
@marktani I don't think using `on-failure` is the way to do so I've just opened an issue instead to keep it in mind https://github.com/prisma/prisma-examples/issues/64.

PR #103, opened 2018-10-03
Demonstrate how a CI setup with CircleCI can look like

w0wka91:


abhiaiyer91:
Awesome PR! Thank you! We were actually thinking of having another CI setup for each individual example that would verify that each example is runnable. Would you be interested in contributing that for one and we can expand to each? Not sure where a stand-alone circle ci example fits in our example list, maybe @schickling can comment more

w0wka91:
Did you think about something like this? https://github.com/w0wka91/typescript-script
I added the circleci config to the typescript-script example and executing it within CircleCI. https://circleci.com/gh/w0wka91/typescript-script/4
There is one problem though. The script doesn't return and therefor the build process runs forever. I think this issue is related to this. https://github.com/prisma/prisma/issues/3226
Im happy to contribute that if that`s what you were thinking of. You could just merge it whenever the abovementioned issue is solved.

abhiaiyer91:
@w0wka91 awesome!

Yes thats very much along the lines of what we were thinking!

Basically a circle ci config per example that:
* Sets up the environment
* Deploys the data model
* Runs prisma generate for the client
and exits.

abhiaiyer91:
@schickling yes definitely this PR is a standalone example no misunderstanding there. My point is that we can use this example as the blueprint to adding CI to each existing and future example and if @w0wka91 is interested in contributing that I’d like to give him the proper guidance

schickling:
I see. I'd actually rather advocate for keeping the examples as minimal/orthogonal as possible. If someone wants to see how testing/CI works, they can take a look at the specific example.

w0wka91:
@schickling I've added all the requested changes. A working example can be found [here](https://github.com/w0wka91/prisma-circleci).
@abhiaiyer91 yes, I'm interested in contributing. :)

typescript-circleci/.circleci/config.yml
```diff
@@ -0,0 +1,28 @@
...
+  build:
+    machine: true
+    working_directory: ~/repo
+    steps:
+      - checkout
+      - run:
+          name: Start Prisma Server
+          command: |
+            set +e             
+            curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.33.11/install.sh | bash
```
schickling: Is there be an easier way to install NodeJS? Ideally in a single line of code.

typescript-circleci/.circleci/config.yml
```diff
@@ -0,0 +1,28 @@
...
+            echo 'export NVM_DIR="/opt/circleci/.nvm"' >> $BASH_ENV
+            echo "[ -s \"$NVM_DIR/nvm.sh\" ] && . \"$NVM_DIR/nvm.sh\"" >> $BASH_ENV
+            cd prisma
+            docker-compose up -d
+            docker ps
+            cd ..
+            npm install
+            cd prisma
+            npx prisma deploy
+            sleep 20
```
schickling: Why is the `sleep` needed?

typescript-circleci/.circleci/config.yml
w0wka91: I tried it now for a few hours but i can't find an easier way.

typescript-circleci/.circleci/config.yml
w0wka91: You're right. This `sleep`is not needed. I remove it from the config file

schickling:
Thanks so much for this great contribution @w0wka91 🙌

PR #110, opened 2018-10-04
Graceful shutdown in Go-GraphQL example

AwesomeObserver:
More extended Go-GraphQL example 😉


schickling:
Thanks a lot for this great PR @AwesomeObserver! 🙏 

While I really appreciate this contribution, I won't merge it for the following reason: The examples are intended to provide a minimal setup instead of a production-ready setup. A production-ready setup often depends on your deployment environment (e.g. AWS, Kubernetes...), so it's hard to draw a line once adding these kinds of features.

I hope this makes sense to you. Thanks again! 👍

PR #131, opened 2018-10-09
Add tests to node-graphql using jest

chakrihacker:


chakrihacker:
Hey, if the example test is fine then, I will continue to write tests for the rest of the examples

divyenduz:
@chakrihacker : Thanks for the PR, looks awesome. Mocking is definitely a very good start and later we can also run Prisma in CircleCI and run it agains a real GraphQL server, up to you if you want to do that right away. This is awesome 🙌

divyenduz:
@abhiaiyer91 : FYI, let us merge and make this awesome together with @chakrihacker :)

abhiaiyer91:
Do we need babel to use jest?

chakrihacker:
Yes, we need babel, since you can't use `import` and other stuff as node still doesn't support **es modules**

abhiaiyer91:
Oh I def understand that, but do we have to add this just for import syntax. Can we write the test in common js?

Pinging @nikolasburk here for his opinion. Originally we had meant to keep the `node` examples as minimal as possible. Not sure how adding babel for the tests here conflicts with that goal

nikolasburk:
Hey @chakrihacker, thanks so much for this PR! This is very much appreciated, thanks a lot for thinking along and actively helping us out 🙏 

As @abhiaiyer91 pointed out, it is a core goal for the examples to be as simple and minimal as possible. I've just added some [Contribution Guidelines](https://github.com/prisma/prisma-examples/blob/master/CONTRIBUTING.md) to this repo which clarify this (this document will be further extended in the future). I think for this particular case, the testing is not part of the core functionality that we want to demonstrate with the `node-graphql` example, so I'd rather not merge it at this point. However, I think it is worth thinking about adding a dedicated example for testing a GraphQL server with Jest to this repo. We're currently still thinking about structuring the examples in this repo, here's what it'll probably look like:

- The top-level examples are the most common use cases for Prisma and will be maintained primarily by the Prisma team
- We'll then add another directory called `community` where we accept examples for more "niche" use cases. I think a `node-graphql-jest` example would be a great fit for that. 

Thanks again for your contribution! 🙌

nikolasburk:
Hey @chakrihacker, I want to apologize again for the confusion. It was clearly on my plate that I wasn't aware of the previous discussions about adding tests to the examples. We first want to work out a coherent strategy for adding tests to examples before we execute on in. We will let you know and I'm really hoping that we can keep you involved, please accept my personal apology for the frustration that closing the PR might have caused on your end 🙏

chakrihacker:
@nikolasburk there is no need of apology.

PR #165, opened 2018-10-13
Go Rest example

abhiaiyer91:
Reviews would be appreciated @divyenduz @AwesomeObserver @Sach97


Sach97:
Maybe we could refactor `w.Header().Set("Content-Type", "application/json")` in a [custom handler](http://www.gorillatoolkit.org/pkg/handlers)

go-rest/main.go
```diff
@@ -0,0 +1,142 @@
...
+				ID: &id,
+			},
+			Data: &prisma.PostUpdateInput{
+				IsPublished: &published,
+			},
+		},
+		).Exec()
+
+		if err != nil {
+			panic(err)
```
AwesomeObserver: Just put this in each error
```go
log.Fatal(err)
```

PR #252, opened 2018-11-13
fix for prisma 1.20.1 when using prisma generate

chrisbull:
When running prisma@1.20.1 and then running `prisma generate` generates an error with the models and `graphqlgen` can't generate the resolvers appropriately. 

The fix was to follow the example in the [Prisma README.md](https://github.com/prisma/graphqlgen), by removing path: and defaultName in the graphqlgen.yml file.

Then when running `prisma generate && graphqlgen` the prisma models build correctly and so does graphqlgen


nikolasburk:
Thanks a lot for this PR and apologies for the delay @chrisbull 🙏 We've addressed this issue in the latest refactoring and cleanup effort of the examples.

PR #255, opened 2018-11-13
add typescript authentication example with graphql-yoga lambda-typescript example

chrisbull:
I wanted to use graphqlgen, prisma-client, and prisma-server along with aws-serverless, so I ended up creating a combination example of typescript-graphql-auth-advanced with [graphql-yoga/examples/lambda-typescript](https://github.com/prisma/graphql-yoga/tree/master/examples/lambda-typescript) and thought I'd share. 

I'm not sure if I did the authentication via context the "recommended way". But it seems to be the only way to read the Authorization from the context. 

Open to suggestions.


nikolasburk:
Hey Chris, thanks a lot for submitting this example and apologies for not having gotten to it! Since we we released Prisma 2, the `prisma-examples` repo is mainly used for Prisma 2 examples and we're not adding any new Prisma 1 examples any more.

That being said, we'll add a new Serverless example for Prisma 2 very soon! Thanks again for your help! 🙏

PR #284, opened 2018-11-25
edit path in README.md

po3rin:
Fixed the document directory path was incorrect.


nikolasburk:
Thanks a lot for catching this @po3rin 🙌

PR #497, opened 2019-03-07
Fix typo

vasco3:


nikolasburk:
Thanks a lot @vasco3. I actually entirely removed the console.log statement since it wasn't supposed to be in there in the first place 🙏

PR #505, opened 2019-03-11
README fix

623841176:
Some teeny bug in graphql-auth and graphql-subscriptions. Field `authorEmail` is not need. And in graphql-subscriptions, there is no need to add `Authorization` header.

### DIFF
--- a/node/graphql-auth/README.md
+++ b/node/graphql-auth/README.md
@@ -184,7 +184,6 @@ mutation {
   createDraft(
     title: "Join the Prisma Slack"
     content: "https://slack.prisma.io"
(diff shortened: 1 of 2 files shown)


623841176:
@nikolasburk I see these are your git commits.

nikolasburk:
Hey @623841176, thanks so much for this PR! 🙏  We actually have a custom templating system that generates the READMEs per example. If you want to fix something, you need to updated the template and re-generate the READMEs. You can find more info about that [here](https://github.com/prisma/prisma-examples/blob/master/CONTRIBUTING.md#improving-a-readme).

I'm happy to merge your PR if you can update it by following this process 🙂

PR #651, opened 2019-07-02
Switch from dep to go mod

matthewmueller:
Switch example from `dep` to `go mod`. The Go authors have deemed `go mod` the way forward, so this updates our example to use that.


nikolasburk:
Thanks Matt 😉

PR #803, opened 2019-09-19
change Photon.default() to Photon()

pantharshit00:
Since we now perform named export, .default will not work and throw an error.

This PR fixes that issue


nikolasburk:
Thanks Harshit 🙌

PR #861, opened 2019-10-11
Fix an old reference to the start command in README

dmzza:
I updated the name of the command to start a server and to disambiguate it from `prisma2 dev` I wrote the whole command `npm run dev`.

### DIFF
--- a/typescript/graphql/README.md
+++ b/typescript/graphql/README.md
@@ -226,7 +226,7 @@ mutation {
 
 To make changes to the GraphQL schema, you need to manipulate the `Query` and `Mutation` types that are defined in [`index.ts`](./src/index.ts). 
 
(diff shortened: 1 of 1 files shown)


nikolasburk:
Hey @dmzza, thanks so much for this PR! Note that we have a "build/templating system" for our example READMEs which means the changes that are made to an example README will be overridden the next time the READMEs are generated. The proper way to fix an issue in the README is by adjusting the README's template in the `.github/readmes` folder and then run `yarn build-readmes` inside `/.github/tests`. You can read more about this here: https://github.com/prisma/prisma-examples/blob/prisma2/CONTRIBUTING.md#improving-a-readme

Nonetheless, thanks a lot for catching this issue and taking the time to submit a PR – I just fixed the issue via our templates and therefore will close this one.

dmzza:
😆 I think I actually started to read the contributing guidelines before I submitted this PR, but I stopped after I read the [README Guidelines](https://github.com/prisma/prisma-examples/blob/prisma2/CONTRIBUTING.md#readme-guidelines) because I knew that's all I was changing. I didn't expect there to be a second section about READMEs at the bottom. 

Thanks for making the fix the right way.

PR #867, opened 2019-10-22
fix: ts nexus use require.resolve sans ext

### LINKED ISSUE #866: Typescript Examples fail after building
Currently, all the typescript examples will fail after being built. This is because of the `require.resolve` call to `context.ts`. After running the build script (tsc), the compiled files will still reference that `context.ts` but that file will not be there since the outDir for typescript is `dist`.

```
Error: Cannot find module './context.ts'
```

The compiled schema.js will have this line

```
source: require.resolve('./context.ts'),
```

This fails because the file is no longer adjacent to `context.ts` so the relative path breaks.

```bash
# Before - Works Typescript (npm run dev)

src
  schema.ts
  context.ts
```

```bash
# After Compile - Fails Javascript

dist
  schema.js # still references ./context.ts
src
  schema.ts
  context.ts
```

All the **typescript examples** break,  but here's just one, showing where the offending line is:
https://github.com/prisma/prisma-examples/blob/prisma2/typescript/graphql-auth/src/schema.ts#L15

### Steps to reproduce
- `npx prisma2 test`
- Select `Starter kit`
- Select `TypeScript`
- Select `Graphql API`
- Select `SQLite`
- After the project is finished installing, run `npm run build`
- Run the compiled output `node dist/server.js`

### Potential fix
This is what i'm using to make it work

```diff
- source: require.resolve('./context.ts'),
+ source: require.resolve('../src/context.ts'),
```

jasonkuhrt:
closes #866

### DIFF
--- a/misc/react-graphql-fullstack/server/package.json
+++ b/misc/react-graphql-fullstack/server/package.json
@@ -1,6 +1,7 @@
 {
   "name": "typescript-graphql",
   "scripts": {
+    "start": "node dist/server",
     "clean": "rm -rf dist",
     "build": "npm -s run clean && npm -s run generate && tsc",
     "generate": "npm -s run generate:prisma && npm -s run generate:nexus",
--- a/misc/react-graphql-fullstack/server/src/schema.ts
+++ b/misc/react-graphql-fullstack/server/src/schema.ts
@@ -103,7 +103,7 @@ export const schema = makeSchema({
         alias: 'photon',
       },
       {
-        source: require.resolve('./context.ts'),
+        source: require.resolve('./context'),
         alias: 'Context',
       },
     ],
--- a/typescript/graphql-apollo-server/package.json
+++ b/typescript/graphql-apollo-server/package.json
@@ -1,6 +1,7 @@
 {
   "name": "typescript-graphql",
   "scripts": {
+    "start": "node dist/server",
     "clean": "rm -rf dist",
     "build": "npm -s run clean && npm -s run generate && tsc",
     "generate": "npm -s run generate:prisma && npm -s run generate:nexus",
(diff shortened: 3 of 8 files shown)


nikolasburk:
Thanks a lot Jason! 🙌

iRoachie:
I think this change resolves to the compiled context.js correct? GraphQL Nexus Typegen needs to resolve context to a typescript file. 

Running this with the change results in:

```
GraphQL Nexus Typegen: Expected module /Users/kyleroach/Desktop/prisma2-heroku/dist/context.js to be an absolute path to a TypeScript module, skipping.
Missing required typegen import: Context
```

jasonkuhrt:
> This change resolves to the compiled context.js now correct?

This change makes it so that a path to a file is returned in both a TS (src) or JS (dist) context.

If you continue to have trouble feel free to open a new issue or reach out on slack. 🙂

iRoachie:
> This change makes it so that a path to a file is returned in both a TS (src) or JS (dist) context.

Not sure if you understood me, the typings for Graphql Nexus require that the source is a path to a TS file with typings. With this change that you made it will point to a JS file. 

Before and after build, the schema file should point to context.ts

See source for typeGen https://github.com/prisma-labs/nexus/blob/9727c0ee86735e054280525cec03ee08658ca02d/src/typegenAutoConfig.ts#L182

jasonkuhrt:
> Before and after build, the schema file should point to context.ts

@iRoachie thanks for the follow up, the problem makes sense. It strikes me as an odd and counter intuitive developer experience though. A revision to the dx around this area is being tracked in https://github.com/prisma-labs/nexus/issues/265.

jasonkuhrt:
Hey @iRoachie, a correction from my previous statement. If you have `NODE_ENV=production` then development mode features should be automatically disabled. Including that LOC you linked to. This is by design. We still think it is too easy to get this wrong, and needs higher-order tooling, but the point right now is, please ensure `NODE_ENV=production`. If it is, and you still see that warning, please open a new issue 🙂

PR #930, opened 2019-11-03
Fix broken link

brunocrosier:
The instructions will fail because https://github.com/prisma/photonjs/tree/master/examples/typescript/codesandbox does not exist anymore. Potentially it might even be better to provide a direct link to codesandbox (https://codesandbox.io/s/github/prisma/prisma-examples/tree/prisma2/deployment-platforms/codesandbox)


nikolasburk:
Thanks a lot for catching this @brunocrosier 🙌

PR #1002, opened 2019-11-20
migrate script examples to facade package

timsuchanek:


divyenduz:
Added a commit with:

1. Removal of `postinstall` hook
2. Using latest alpha of `@prisma/photon` and `prisma2`

Pending: 

-[ ] Use latest nexus in the GraphQL examples

PR #1035, opened 2019-11-26
Fix query argument in example

2color:


2color:
@nikolasburk Thanks for the heads-up. I've updated the PR accordingly.

nikolasburk:
Amazing, thank you so much! 🙏

PR #1052, opened 2019-11-27
Set NODE_ENV to 'production' in 'netlify.toml'

kripod:
I was having issues with setting up Netlify Functions, when I realized that environment variables shall be configured prior to deployment. By adding these lines, the environment will be set to `'production'` out of the box.


nikolasburk:
Thanks a lot for catching this @kripod and apologies for the delay! 🙏

PR #1053, opened 2019-11-27
fix: update gql examples

jasonkuhrt:


nikolasburk:
Thanks Jason, I think this was fixed in the meantime 🙏

PR #1160, opened 2019-12-12
Add a GraphQL API example with a union type built with Typescript and Photon.js

chief-wizard:
This new example is based on [an existing GraphQL API example from this repo](https://github.com/prisma/prisma-examples/tree/prisma2/typescript/graphql).

The main differences vs. the [GraphQL API example](https://github.com/prisma/prisma-examples/tree/prisma2/typescript/graphql) are:

1. `Post` is a union type. These resolvers use the union `Post` type and return both articles and photos:

* `feed`
* `filterPosts`

2. `User` is an interface, we can either have Authors or Photographers, and the idea is that Authors only have articles associated with them and Photographers only have photos.


nikolasburk:
We never got around polishing this example to publish it so I'll close this PR. If we decide to add a similar example we can use this one for reference.

(And cheers to @chief-wizard for your work here, hope you're doing fine Alexey 🙂 )

chief-wizard:
Sounds good — thanks @nikolasburk and hope you are well too!

PR #1408, opened 2020-02-08
fix(docs): Use global Express.JS website link instead of German one

janpio:
Replace all German Express website links with global one.

Additionally Express calls itself just "Express", not "Express.JS" - so this would probably be a good PR to also change that if wanted.


nikolasburk:
Thanks for catchinh that Jan 🙌

PR #1409, opened 2020-02-08
fix(rest-express): sample requests link was broken

janpio:
Might also have to be applied to other files. Feel free to edit the PR to do that.


nikolasburk:
Thanks Jan, fixed in both the JS and the TS example 🙏

PR #1448, opened 2020-02-18
chore: add project-specific .gitignore files

### LINKED ISSUE #1423: Create project-specific .gitignore files
I think we should have a ready to use .gitignore files in each project, so `node_modules` are automatically ignored when the user clones the repo and `cd`s into a project. That also allows us to ignore project-specific generated files in our repository and it will behave the same way when a user uses an example to start a project of their own.

steebchen:
fix #1423


steebchen:
Done!

nikolasburk:
Thanks Luca!! 🎉

PR #1455, opened 2020-02-18
Replace @prisma/photon with @prisma/client

marcjulian:


nikolasburk:
Thanks a lot for catching this @marcjulian, I guess the example still won't be functional since we also need to adjust the naming in the code. We'll update the deployment examples soon! 🙏

PR #1463, opened 2020-02-19
ci(circleci): remove CI

steebchen:
prisma 1 is in maintenance mode and the current CI fails and is unmaintained as well


nikolasburk:
Thanks 👍

PR #1487, opened 2020-02-21
ci(github): initialize per-project tests

### LINKED ISSUE #1439: Set up basic tests in CI
We should make sure our examples have small basic tests to make sure they compile and are runnable. This should be as minimal as possible, since we already cover testing various tools in https://github.com/prisma/prisma2-e2e-tests.

steebchen:
fix #1439


janpio:
No individual checks per project? By choice or design limitation?

steebchen:
The advantage of having everything one one check is that all projects are tested per default, and new projects without CI fail per default. In contrast, when adding new projects, it also has to be manually and separately added to CI. It also may overwhelm users when your PR needs to have 35 checks. We can totally do that, I don't care so much as I'm fine with both approaches.

nikolasburk:
Thanks a lot Luca, that's great! 🙏

PR #1638, opened 2020-03-25
fix isPostOwner

### LINKED ISSUE #1637: isPostOwner in permissions file for graphql-shield is wrong
Trying to run the default auth examples result in an error when calling the isPostOwner shield rule.

It currently looks like this
```javascript
isPostOwner: rule()(async (parent, { id }, context) => {
    const userId = getUserId(context)
    const author = await context.prisma
      .post({
        id,
      })
      .author()
    return userId === author.id
  }),
```

The problem is that `ctx.prisma.post()` is not a function, I am pretty sure this is a leftover from prisma1 but I just started using prisma and version 2 is the first one I use so I am not sure.
I managed to fix the problem with the following code (after pulling my hair out for way longer than I should have)
```javascript
isPostOwner: rule()(async (parent, { id }, context) => {
    const userId = getUserId(context)
    const author = await context.prisma
      .post.findOne({
        where: {
          id: parseInt(id, 10)
        }
      })
      .author()
    return userId === author.id
  }),
```

I have made a corresponding PR if any maintainer has the time to look over it. The problem is probably also present in the typescript version and every example where auth might be used.

jukefr:
fix #1637

### DIFF
--- a/javascript/graphql-auth/src/permissions/index.js
+++ b/javascript/graphql-auth/src/permissions/index.js
@@ -6,11 +6,13 @@ const rules = {
     const userId = getUserId(context)
     return Boolean(userId)
   }),
-  isPostOwner: rule()(async (parent, { id }, context) => {
+isPostOwner: rule()(async (parent, { id }, context) => {
     const userId = getUserId(context)
     const author = await context.prisma
-      .post({
-        id,
+      .post.findOne({
+        where: {
+          id: parseInt(id, 10)
+        }
       })
       .author()
     return userId === author.id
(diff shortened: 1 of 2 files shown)


nikolasburk:
Hey @jukefr, thanks so much for this PR 🎉  I would indeed ask you to replace `parseInt` with `Number` to remain consistent across all examples. 

Also, I believe that there probably is the same issue in the `typescript/graphql-auth` example, can I ask you to fix it there as well (if needed)?

Thanks a lot, looking forward to merging this! 🙂

jukefr:
The typescript one actually had the right way to call post.findOne
I still added Number to explicitly typecast id

nikolasburk:
Thanks so much @jukefr 🎉

PR #1644, opened 2020-03-27
Add aws lambda deployment example

2color:


2color:
Merging following chat with @nikolasburk

Issue #1613, opened 2020-03-17
Remove `dom` and add `@types/node` in TS examples


nikolasburk:
Context here: https://github.com/prisma/prisma2/pull/1858

janpio:
Is this still open after the PR has been merged @nikolasburk?

nikolasburk:
Yup, seems like this has been removed from all examples in favor of adding `@types/node` 👍

PR #1647, opened 2020-03-27
Remove dom from tscongif + add types/node

createvo:
Added @types/node where missing and removed dom lib in TSconfig. 
I'm not sure if it can be done also for "nextjs" and "typegraphql" examples. 
#1613


nikolasburk:
Thanks a lot for taking this over @createvo 🚀

PR #1650, opened 2020-03-28
Add experimental react-graphql-fullstack back

janpio:
Was removed because tests were failing. Needs to be fixed before can be merged again.


steebchen:
Note: @prisma/cli is missing here, which is why it fails at upgrading

nikolasburk:
This example was not well maintained, I think we can get rid of it altogether

PR #1681, opened 2020-03-31
Fix prisma versions to beta in starters

pimeys:


nikolasburk:
Thanks so much for helping out here Julius 🚀

PR #1692, opened 2020-04-01
Update TypeGraphQL domain in READMEs

MichalLytek:
The domain has changed a few weeks ago from `typegraphql.ml` to `typegraphql.com`


janpio:
Thanks @MichalLytek

PR #1702, opened 2020-04-03
Remove unused dependencies and fixed Post component error

shaswatsaxena:
- Removed unused dependencies: express and tachyons from both JS and TS examples
- Post.jsx component fixed in JS rest-nextjs example to fix nested a tags error.


nikolasburk:
🚀 🚀 🚀

PR #1714, opened 2020-04-09
Fix searchString missing type

devalnor:
This fix the typescript error:

Type 'Query' is not assignable to type 'string'.

Detail:
src/index.ts:76:13 - error TS2322: Type 'string | Query | string[] | Query[]' is not assignable to type 'string | null | undefined'.
  Type 'Query' is not assignable to type 'string'.

76             contains: searchString,
               ~~~~~~~~
src/index.ts:81:13 - error TS2322: Type 'string | Query | string[] | Query[]' is not assignable to type 'string | null | undefined'.
  Type 'Query' is not assignable to type 'string'.

81             contains: searchString,
               ~~~~~~~~

`contains: <string>searchString` or `contains: searchString as string` are also possible solutions but less cleaner.

### DIFF
--- a/typescript/rest-express/src/index.ts
+++ b/typescript/rest-express/src/index.ts
@@ -67,7 +67,7 @@ app.get('/feed', async (req, res) => {
 })
 
 app.get('/filterPosts', async (req, res) => {
-  const { searchString } = req.query
+  const { searchString }: { searchString?: string } = req.query;
   const draftPosts = await prisma.post.findMany({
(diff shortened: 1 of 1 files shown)


nikolasburk:
Thanks a lot for catching this @devalnor 🚀

desy83:
@devalnor why is it necessary to set the type definitions for request query now? was there an update of node or typescript? thanks

PR #1715, opened 2020-04-09
ci(test): trigger on pull_request and master

steebchen:


janpio:
Is there a way to not have 2 workflows running the tests if we ourselves open a Pull Requests for a branch in this repo?

steebchen:
I don't think so, that's why I didn't merge yet

.github/workflows/test.yaml
```diff
@@ -1,5 +1,12 @@
 name: test
-on: push
+on:
+  push:
+    branches:
+      - master
+      - alpha
+      - dev
+      - patch-dev
+  pull_request:
```
janpio, line 9: is this correct with the colon but not sub value?

steebchen: yep

steebchen: seems a bit weird, but works indeed

PR #1749, opened 2020-04-25
[rest-express] Parse id from req.params

### LINKED ISSUE #1748: Express example throws on specific routes
These examples are very helpful in seeing Prisma in action but I found that the Express example didn't resolve any of the default routes with `id` params. It turns out the id from `const { id } = req.params` expects an integer in the defined schema but what it receives a string. 

Instead of using object destructuring for `id` it would work if `id` was defined with `parseInt(id)`.

josephdburdick:
Fixes #1748


nikolasburk:
Thanks a lot for catching this @josephdburdick 🙏

PR #1755, opened 2020-05-04
Fix build for typescript/graphql-nextjs

akshaylive:
1. `generate` script is broken due to reliance of `commonjs` by nexus. Fixing this by changing module to `commonjs`. One issue with this approach is that the next time `yarn dev` is run, `nextjs` automatically identifies the module issue and fixes it to `esnext`. This behavior is preferred over breaking builds during `nexus` generation.
1. `__dirname` doesn't seem to exist (tested on Ubuntu). Instead, relying on `process.cwd()`, and building the path from there. This will ensure that the code remains platform agnostic.
1. Updating dependencies. This fixes the broken `build` script.
1. `createdAt` and `udpatedAt` is not part of the `Post` schema, and hence should not be defined on nexus object type.
1. `id` is defined as an `Int` in the schema. Hence, the nexus type for `id` should be `int`.


nikolasburk:
Thanks so much for cleaning all this up @akshaylive 🚀 🙏

PR #1761, opened 2020-05-08
Fix README

tconroy:
this PR fixes an incorrect type for the `User` reference in the Profile SQL example.

Previously, User ID was being referenced as `TEXT`, when it should be `INTEGER`.


nikolasburk:
Thanks a lot for fixing this @tconroy 🚀

PR #1767, opened 2020-05-11
improve(workflows): rename workflows to clearer names

janpio:
This renames the `check-for-update` GH Action workflow to `keep-prisma-dependencies-updated`, and `sync-alpha` to `keep-alpha-branch-in-sync-with-master` to better highlight what they and why it is important that they succeed.

(The badge in the README should work again after this is merged and the workflow ran once)


janpio:
Let's see if @steebchen agrees, who built the thing 👍

steebchen:
We might want to rename the workflow & script files as well, though they would get quite long

PR #1768, opened 2020-05-11
docs(README): add gh action badges

janpio:
This PR adds the two important GH actions as badges to the README to increase visibility into possible failures.

Open for better positioning or styling of course, this is a bit "brute force".


nikolasburk:
Nice, thanks a lot Jan!

PR #1802, opened 2020-06-03
Improve Vercel GraphQL example

2color:
## What
- Upgrade `nexus-prisma` -> nexus-plugin-prisma/schema
- Make fields required in queries
- Add users queries
- Add better dev command

### DIFF
--- a/deployment-platforms/vercel-graphql/api/_lib/context.ts
+++ b/deployment-platforms/vercel-graphql/api/_lib/context.ts
@@ -1,7 +1,8 @@
 import { PrismaClient } from '@prisma/client'
 
 const prisma = new PrismaClient({
-  log: ['query', 'info', 'warn'],
+  // Uncomment for debugging purposes
+  // log: ['query', 'info', 'warn'],
 })
 
 export interface Context {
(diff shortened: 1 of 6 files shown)


deployment-platforms/vercel-graphql/api/_lib/context.ts
```diff
@@ -1,6 +1,8 @@
 import { PrismaClient } from '@prisma/client'
 
-const prisma = new PrismaClient()
+const prisma = new PrismaClient({
+  log: ['query', 'info', 'warn'],
```
divyenduz, line 4: Do we want to keep this logging in the examples?

2color: Yes.

2color: My thought was that this provides useful insight into what's happening with the application as queries come in.

Jolg42: I would put it commented with a comment "Useful for debugging..." or remove it personally

PR #1830, opened 2020-06-23
Rename relation fields and add renamed schema

2color:


2color:
> I'm wondering why we chose snake case over camelCase for the field names; afaik we usually prefer camelCase. otherwise lgtm

For historical reasons. This fix is mostly to get the example to a working state. I'm happy to update it to camelCase next week.

PR #1838, opened 2020-07-02
ci(keep-prisma-dependencies-updated): rename jobs

janpio:
and add patch-dev as needs for slack failure report as well


.github/workflows/keep-prisma-dependencies-updated.yaml
```diff
@@ -3,7 +3,7 @@ on:
   schedule:
     - cron: '*/5 * * * *'
 jobs:
-  check-for-preview-update:
+  check-for-update-latest:
```
janpio, line 6: this was the important one - no more preview.

PR #1844, opened 2020-07-03
feat(postgis): Add new example with PostGIS

ryands17:


ryands17:
@janpio I can add the postgres service to the main test and then run it in a single workflow it but that would require all the other tests to wait for the database to spin up. Would that be fine?

janpio:
Don't other tests also use a (remote) Postgres database? What is different about this example that it needs to be spun up just for the test?

ryands17:
@janpio All the tests use SQLite. This is the first Postgres example :)

janpio:
@steebchen In e2e tests we use persistent Postgres databases from outside of the repo context. Do you have a preference what is cleaner here - do that as well or add Postgres to the workflow?

steebchen:
I think if possible it's best to not depend on any external service and set up everything in CI ("locally"). We only do that in e2e because most platforms already depend on third party services, so it really doesn't make a difference. For examples, AFAIK we only use local tests, so I think adding a local Postgres is best.

janpio:
> but that would require all the other tests to wait for the database to spin up.

Is that actually a problem?
Is there maybe a way to only do this for a specific example?

steebchen:
> Is that actually a problem?

No. The tests are all blocking anyway.

> Is there maybe a way to only do this for a specific example?
It is when we define a separate action which seems is the case already, but it could just as well be set up in the other tests. However, it's probably one or the other – either we do everything in one check, or we do one check per test (but they _will_ add up).

janpio:
> It is when we define a separate action which seems is the case already, but it could just as well be set up in the other tests. However, it's probably one or the other – either we do everything in one check, or we do one check per test (but they _will_ add up).

I blocked this PR as it introduces an additional workflow besides the standard one to run the tests as this will lead to drift between this implementation and the main test implementation.

Are you referring here to the option to similarly structure this repo's checks like in the e2e repo?

steebchen:
> Are you referring here to the option to similarly structure this repo's checks like in the e2e repo?

Yes.

janpio:
I will put this into a new issue.

For @ryands17 the right thing to do for now would be to combine his new workflow into the existing one to make sure we keep the current test structure for now, correct @steebchen?

ryands17:
@janpio changes done but the master branch has breaking tests, maybe due to a version bump of a `graphql-yoga`.

.github/readmes/typescript/postgis-express/README.md
```diff
@@ -0,0 +1,25 @@
+# Postgres Geolocation example
+
+This example shows how to implement **Geolocation in Postgres** using [Express](https://expressjs.com/), [Postgis](http://postgis.net/) and [Prisma Client](https://github.com/prisma/prisma2/blob/master/docs/prisma-client-js/api.md).
```
janpio, line 3:
```suggestion
This example shows how to implement **Geolocation in PostgreSQL** using [Express](https://expressjs.com/), [Postgis](http://postgis.net/) and [Prisma Client](https://github.com/prisma/prisma2/blob/master/docs/prisma-client-js/api.md).
```

Is the "Geolocation" the correct label for this? Or maybe better "Geolocation Types" or similar?

.github/readmes/typescript/postgis-express/_setup-postgres.md
```diff
@@ -0,0 +1,28 @@
+### 2. Setup Postgres
```
janpio, line 1:
```suggestion
### 2. Setup PostgreSQL
```

.github/readmes/typescript/postgis-express/_setup-postgres.md
```diff
@@ -0,0 +1,28 @@
...
+
+- The recommended way of trying this out would be using a Docker image of Postgres + Postgis that can be pulled from [here](https://github.com/postgis/docker-postgis).
```
janpio, line 3:
```suggestion
- The recommended way of trying this out would be using a Docker image of PostgreSQL with Postgis extensions that can be pulled from [here](https://github.com/postgis/docker-postgis).
```

.github/readmes/typescript/postgis-express/_setup-postgres.md
```diff
@@ -0,0 +1,28 @@
...
+```
+
+- Lastly run the following commands to introspect the database and generate Prisma Client.
+
+```
+npm run prisma -- introspect
+npm run prisma -- generate
+```
+
+__*Limitation*__: Currently Prisma doesn't support custom data types so querying for the *geography* type is not possible. Only operations can be performed on the types via `prisma.queryRaw` or `prisma.executeRaw`.
```
janpio, line 28:
```suggestion
__*Limitation*__: Currently Prisma doesn't support custom data types, so querying for the *geography* type in the normal Prisma models is not possible. Operations can only be performed on the types via `prisma.queryRaw` or `prisma.executeRaw`.
```

The raw command could possibly be linked to the docs.

.github/workflows/test.yaml
```diff
@@ -3,14 +3,28 @@ on: push
 jobs:
   test:
     runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v1
+    container: node:12.18.2-stretch
```
janpio, line 6: why the explicit definition of a container here?

.github/workflows/test.yaml
```diff
@@ -3,14 +3,28 @@ on: push
...
-      - uses: actions/setup-node@v1
-        with:
-          version: 12
+    steps:
+      - uses: actions/checkout@v2
 
       - name: test
         run: sh .github/scripts/test-all.sh
         env:
+          CI: true
```
janpio, line 29: why is this necessary manually?

typescript/postgis-express/package.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "postgis-express",
+  "version": "1.0.0",
+  "license": "MIT",
```
janpio, line 4: Is MIT the correct licence for our examples?

janpio:
Intentional this was merged without adding it to https://github.com/prisma/prisma-examples/blob/master/README.md?

ryands17:
@janpio I'll add both in the testing example itself.

PR #1853, opened 2020-07-07
fix(graphql-auth): remove projects as tests are failing

janpio:


steebchen:
Can't reproduce locally...

PR #1859, opened 2020-07-09
Feature/testing

ryands17:


deployment-platforms/aws-lambda/package.json
```diff
@@ -12,5 +12,5 @@
     "@prisma/cli": "2.2.0",
     "serverless-dotenv-plugin": "2.4.2"
   },
-  "license": "MIT"
+  "license": "Apache 2.0"
```
janpio, line 15: If we are doing this, it should be done in a separate PR please that can be merged on its own.

PR #1940, opened 2020-08-09
fix link

gregbenner:
fix link

### DIFF
--- a/javascript/graphql-sdl-first/README.md
+++ b/javascript/graphql-sdl-first/README.md
@@ -33,7 +33,7 @@ Navigate to [http://localhost:4000](http://localhost:4000) in your browser to ex
 
 ## Using the GraphQL API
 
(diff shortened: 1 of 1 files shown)


javascript/graphql-sdl-first/README.md
```diff
@@ -33,7 +33,7 @@ Navigate to [http://localhost:4000](http://localhost:4000) in your browser to ex
 
 ## Using the GraphQL API
 
-The schema that specifies the API operations of your GraphQL server is defined in [`./schema.graphql`](./schema.graphql). Below are a number of operations that you can send to the API using the GraphQL Playground.
+The schema that specifies the API operations of your GraphQL server is defined in [`./prisma/schema.graphql`](./schema.graphql). Below are a number of operations that you can send to the API using the GraphQL Playground.
```
janpio, line 36: That file also does not exist in this repo.

gregbenner: my bad, maybe the link should just be removed rather than broken

janpio: I agree - or it should be mentioned that/if the file is generated later or during installation. @nikolasburk

PR #1986, opened 2020-09-03
Remove the introspected schema from the example

2color:
Since the schema will be filled by introspection, having it in the repository is not necessary. Moreover, in the [deployment guide](https://www.prisma.io/docs/guides/deployment/deploying-to-vercel#5-introspect-the-database), the checkpoint relies on the user successfully introspecting the database, so having it already there can confuse users.

### DIFF
--- a/deployment-platforms/vercel/prisma/schema.prisma
+++ b/deployment-platforms/vercel/prisma/schema.prisma
@@ -7,25 +7,3 @@ datasource db {
   url      = env("DATABASE_URL")
 }
 
-model Post {
-  post_id   Int     @default(autoincrement()) @id
-  author_id Int?
-  content   String?
-  title     String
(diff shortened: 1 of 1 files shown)


janpio:
> Not sure how useful it is because the comment would probably not be overwritten.

A free floating comment would indeed just go away on ùsing `introspect`.

janpio (commented):
If the example is not runnable by default, and needs manual import and `introspect` of the schema, I am not really sure any more if this should be in this repository, but it really "just" the basis for a guide.

2color:
> If the example is not runnable by default, and needs manual import and introspect of the schema, I am not really sure any more if this should be in this repository, but it really "just" the basis for a guide.

That was the main reason I left it. So I'll just close this for now

PR #2010, opened 2020-09-12
ci(check-for-update): simplify parameters

steebchen:


janpio (approved):
Also updated the usage of the script.

steebchen:
Thanks, somehow I forgot to commit that file...

PR #2011, opened 2020-09-12
ci: simplify keep-prisma-dependencies-updated

janpio:
In theory this could work. But in practice? No idea :/


divyenduz (approved):
Should be fine if the three jobs had no other diff than `${{ matrix.branch }}`

Squinted a bit and looks fine to me ✅

janpio:
Feel free to merge and observe if things still work then.

PR #2012, opened 2020-09-12
ci: tests as matrix

janpio:
Changes the tests to run via a Matrix that is automatically generated from the existing project folders.

Tasks:
- [x] Generate matrix
- [x] Change Shell script to run for 1 project, not all
- [x] Move skipping of folders from script to matrix generation
- [x] Move more from Shell script into step
- [ ] Slack reporting for step replacing shell script
- [ ] Find better names for jobs

Follow up tasks:
- Put Slack reporting into step as well, similar to https://github.com/prisma/language-tools/blob/0b12cb90720a261ebe540ef4dce8edb7c39982e8/.github/workflows/7_publish.yml#L72-L85

### DIFF
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -2,6 +2,7 @@
 
 set -eu
 
+item="$1"
 dir="$(pwd)"
 
 
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -18,7 +18,7 @@ echo "running $item"
 case "$item" in
   *".github"*|*"experimental"*|*"deployment-platforms"*)
     echo "ignoring $item"
-    continue
+    exit 0
     ;;
 esac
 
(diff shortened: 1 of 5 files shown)


.github/scripts/test.sh
```diff
@@ -0,0 +1,77 @@
...
+echo ""
+echo "---------------------"
+echo "running $item"
+
+case "$item" in
+  *".github"*|*"experimental"*|*"deployment-platforms"*)
+    echo "ignoring $item"
+    continue
+    ;;
+esac
```
janpio: TODO: Put this into first step as well, just filter out of list of projects

divyenduz:
This was done via https://github.com/prisma/prisma-examples/pull/2180

Remaining issues are tracked here https://github.com/prisma/prisma-examples/issues/2187 and https://github.com/prisma/prisma-examples/issues/2181

PR #2034, opened 2020-09-23
fix(prisma-aws-lambda+varcel): logical error and code layout fix

rahul3v:


nikolasburk:
Hey @rahul3v, thank you so much for fixing this! Unfortunately, the PR can't yet be merged since we're following the standard of [semantic commits](https://gist.github.com/joshbuchea/6f47e86d2510bce28f8e7f42ae84c716) in this repo, so you'll need to change the commit message to adhere to that format, e.g.: `chore(deployment-platforms/aws lambda: logical error and code layout fix`). I'll be more than happy to merge the PR once that's done 🙂

rahul3v:
@nikolasburk , have tried many possible cases, still not accepting it, should i make it more genral

janpio:
The commit title does not actually matter, as we can just "Squash and Merge" which should use the PR title for the new commit message. But seems you fixed it now as well, so this is good to go.

nikolasburk:
Thanks a lot @rahul3v 🎉

PR #2066, opened 2020-10-05
chores: improve setup

divyenduz:


.github/scripts/test-all.sh
```diff
@@ -17,7 +17,7 @@ echo "$packages" | tr ' ' '\n' | while read -r item; do
   echo "running $item"
 
   case "$item" in
-    *".github"*|*"experimental"*|*"deployment-platforms"*|"typescript/subscriptions-pubsub")
+    *".github"*|*"experimental"*|*"deployment-platforms"*|*"typescript/subscriptions-pubsub"*|*"typescript/postgis-express"*|*"typescript/graphql"*|*"typescript/graphql-nextjs"*|*"javascript/graphql"*|*"javascript/graphql-apollo-server"*|*"javascript/graphql-auth"*)
```
janpio, line 20: What is the reasoning behind ignoring all these?

.github/tests/javascript/graphql-auth/run.sh
```diff
@@ -9,4 +9,4 @@ sleep 15
 
 curl --fail 'http://localhost:4000/' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://localhost:4000' --data-binary '{"query":"query {\n  feed {\n    id\n    content\n    author {\n      id\n      name\n      email\n    }\n  }\n}"}' --compressed
 
-kill "$pid"
+kill "$pid" || true
```
janpio, line 12: Why should this ever not work?

docker-compose.yml
```diff
@@ -0,0 +1,15 @@
...
+    ports:
+      - '5432:5432'
+    environment:
+      POSTGRES_USER: postgres
+      POSTGRES_PASSWORD: password
+      POSTGRES_DB: testing
+    volumes:
+      - postgres:/var/lib/postgresql/data
+volumes:
+  postgres:
```
janpio, line 15: Why is this needed?

typescript/graphql-apollo-server/schema.graphql
```diff
@@ -3,10 +3,8 @@
 
 
 type Mutation {
-  createDraft(authorEmail: String, content: String, title: String!): Post!
-  deleteOnePost(where: PostWhereUniqueInput!): Post
+  createDraft(authorEmail: String!, content: String, title: String!): Post!
   publish(id: Int): Post
-  signupUser(data: UserCreateInput!): User!
```
janpio, line 9: What are all these deletions in this file?

.github/scripts/test-all.sh
divyenduz: `*".github"*|*"experimental"*|*"deployment-platforms"*` I don't know, that is how it was when I picked it. 

Rest use `nexus-prisma` which is broken at the moment.

.github/tests/javascript/graphql-auth/run.sh
divyenduz: No idea but it failed on my local once and the existing code has a catch-all to clean up. This is just a precautionary measure.

docker-compose.yml
divyenduz: For easier local setup. The Pg in the tests comes from actions service but when I want to set it up locally I need to translate that to a `docker-compose.yml` file.

typescript/graphql-apollo-server/schema.graphql
divyenduz: I didn't make the change, it is most likely a "generated" change

.github/scripts/test-all.sh
nikolasburk: Actually, we can keep those after all since we "fixed" the corresponding tests by reverting the versions of the `@prisma` packages back to `2.7.0`. Let's rather keep them in the tests and resolve the issues by reverting to a lower version rather than removing tests. (Sorry I think I wasn't clear about this before).

divyenduz: Noted, will do so in a new PR.

.github/tests/javascript/graphql-auth/run.sh
janpio: Please add an issue to investigate and understand this (or replace that code).

docker-compose.yml
janpio: Root of the repository is not an appropriate location for this then as it gives a wrong impression about what these examples need or want to be run.

typescript/graphql-apollo-server/schema.graphql
janpio: This should be understood then before committing - it changes the example and might be wanted - or not.

divyenduz: `"generate:nexus": "ts-node --transpile-only src/schema"` this is in the `package.json` file, I think this is generated. I call @nikolasburk to validate.

docker-compose.yml
divyenduz: Removed in https://github.com/prisma/prisma-examples/pull/2068

.github/tests/javascript/graphql-auth/run.sh
divyenduz: Removed in https://github.com/prisma/prisma-examples/pull/2068

.github/scripts/test-all.sh
divyenduz: Removed all the nexus ones via https://github.com/prisma/prisma-examples/pull/2079

The "postgis" one is most likely something else (I need to investigate). Will remove that in a separate PR.

Issue #1659, opened 2020-03-28
Tests should report failure if 1 of the tests fails


janpio:
Right now Actions reports the tests are succeeding, although they are clearly failing:

https://github.com/prisma/prisma-examples/issues/1652
https://github.com/prisma/prisma-examples/issues/1651

That is dangerous, as one does not really see when stuff goes wrong.

Fundamentally there are 2 problems:
1. All tests are run in one batch, a suggestion to change that is at https://github.com/prisma/prisma-examples/issues/1653
2. Even when one of the tests clearly fails, the test run still reports a success as a whole.

If one of the tests fails, the run should report a failure as a whole.

steebchen:
The problem is within the tests themselves, not in the general test scheme. The errors you see in the log output have different reasons: one unfortunate mistake from me is misunderstanding Promises which is fixed via https://github.com/prisma/prisma-examples/pull/1661; most other errors are more like warnings and don't have any influence. For example, for graphql test, we send a GraphQL request using curl so if that succeeds it doesn't matter what the server reports; as long as the request succeeds without an error it's fine, but of course it would be much better if we would actually test for the result – which we can of course do but is more work.

janpio:
Right now the situation is that I get a passing workflow. Looking into the output of that, I see all kinds of errors - both real errors and not real errors. In https://github.com/prisma/prisma-examples/issues/1651 the stuff might be cosmetic an what you describe, but https://github.com/prisma/prisma-examples/issues/1652 many tests are really totally broken.

That is what we need to fix.

steebchen:
Can this be closed?

Issue #1653, opened 2020-03-28
Change the way tests are run


janpio:
Right now there is only 1 test, that loops through all folders. That is practical, because it makes sure that all examples (= folders) have the proper test configuration that then is run during the test - no one can add a test without it also being tested.

Unfortunately that leads to super busy log output and makes it hard to understand what is going on.

An alternative suggestion: The test starts with a script that tests if for all folders there is a Github Actions workflow/job that triggers the .sh file that currently would be triggered via looping through the folders. That way, the whole test run can still be failed if an example is missing.

steebchen:
It's probably a bit tricky to test for if there's a workflow defined for a given file. Definitely possible but we probably need to use a yaml parser to read the workflow files properly and then compare stuff, unless you have a better suggestion.

mohe2015:
@steebchen Depending on how the workflow files should look you could also write a script that loops through the folders and just generates a workflow for them. So no parsing just regenerating the file every time.

janpio:
Couldn't it be easy enough as checking if there is a matching `run.sh` being called via its own job to each folder? Instead of looping through the folders and executing the `run.sh`, it would just make sure there is a job defined that `sh ...run.sh` that file.

steebchen:
(Just to clarify, the script already tests if all projects have a run.sh file, and throws if there's a project without a defined test. It's just doing that in the same single test run.)

So did I understand correctly that you want to have one job ("GitHub check") which makes sure all projects have defined a GH workflow (so that script would read .github/workflows/ yaml files and compare strings etc.), and then a dedicated check for each project? Or do you want to have a single job ("GitHub check") as is but split the steps into 1) making sure all projects have test files defined and 2) one separate step for each project?

janpio:
> So did I understand correctly that you want to have one job ("GitHub check") which makes sure all projects have defined a GH workflow (so that script would read .github/workflows/ yaml files and compare strings etc.), and then a dedicated check for each project? 

Yes, that is probably what I meant (if "Job" and "check" are actually the same thing). Basically this changes the testing setup here from 1 job that checks if all projects have tests first, then runs all of them one after another to 1 job that checks if all projects have a job defined and can fail the build if not, and many individual jobs for all projects.

> Or do you want to have a single job ("GitHub check") as is but split the steps into 1) making sure all projects have test files defined and 2) one separate step for each project?

No, I think just switching over to steps would give not much benefit.

PR #2100, opened 2020-10-12
fix: slack notifications

### LINKED ISSUE #2048: Failures are not being reported to slack all the time
`dev` branch is currently failing (expected nexus error) but there are no reports in slack. 

```
TypeError: Cannot read property 'some' of undefined
    at transformArg (/home/runner/work/prisma-examples/prisma-examples/typescript/subscriptions-pubsub/node_modules/nexus-prisma/src/dmmf/transformer.ts:82:33)
```

https://github.com/prisma/prisma-examples/runs/1182310267

Note that it does work sometimes: https://prisma-company.slack.com/archives/CUH6B8FUJ/p1601459380023500

divyenduz:
Fixes https://github.com/prisma/prisma-examples/issues/2048

There is a notification coming from inside `test-all.sh` and that has the context of which test failed. Since we don't have that context outside, we can't yet remove it from `test-all.sh`

This PR adds a catch-all notification without the name of the actual test that fails (so there may be 2 error messages, for now, one with and one without the actual example name). If that is okay, we can already merge this to not miss errors OR we can simply work on making the whole setup better. 

This is properly fixed when examples are being tested in their own step instead of one big step that runs all the tests. 

Related issues/PRs: 

- https://github.com/prisma/prisma-examples/issues/1653
- https://github.com/prisma/prisma-examples/issues/1659
- https://github.com/prisma/prisma-examples/pull/2012

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -7,6 +7,7 @@ on:
       - patch-dev
   pull_request:
 env:
+  CI: 1
   PRISMA_TELEMETRY_INFORMATION: 'prisma-examples test.yaml'
 jobs:
   test:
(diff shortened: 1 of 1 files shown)


.github/workflows/test.yaml
```diff
@@ -7,6 +7,7 @@ on:
       - patch-dev
   pull_request:
 env:
+  CI: 1
```
janpio, line 10: Not necessary in theory as GH Actions should be picked up automatically by the library we are using.

janpio (approved):

.github/workflows/test.yaml
divyenduz: Did that just for consistency with other places where we have `PRISMA_TELEMETRY_INFORMATION `

PR #2111, opened 2020-10-16
feat: slack notif in sync script

divyenduz:


.github/workflows/keep-dev-branch-in-sync-with-latest.yaml
```diff
@@ -27,3 +27,19 @@ jobs:
         run: sh .github/scripts/sync.sh patch-dev
         env:
           SSH_KEY: ${{ secrets.SSH_KEY }}
+
+  report-to-slack-failure:
+    if: github.repository_owner == 'prisma' && failure()
```
divyenduz, line 32: `github.repository_owner == 'prisma'` not sure why is this condition required, It was there in the other workflow but I don't know yet when the repo owner is not `prisma`

janpio (approved):

PR #2188, opened 2020-10-30
feat(deps): bump Prisma deps in latest

### LINKED ISSUE #2063: `@prisma` deps should only be updated if tests succeed
Right now, the sync script that updates the `@prisma` dependencies in all example projects always runs upon new releases of Prisma. 

We often had the situation (and are experiencing it right now again as of the latest 2.8.0 release), that changes in the DMMF are breaking the `nexus-prisma` package which is used in a number of examples. Whenever this happens, the examples don't work any more and the tests for them start failing,

Functional example projects are really important for us, so we should make sure that our sync script only updates Prisma dependencies if the tests for a project succeed.

divyenduz:
Fixes https://github.com/prisma/prisma-examples/issues/2063

This PR 

- Add renovate package rule to bump major versions of `@prisma/cli` and `@prisma/client` in `latest` branch
- Removed the stale code in our custom bump script that ignored Prisma dependencies in `nexus-prisma-plugin` projects
- Disables `latest` in our custom bump script

Note: at a later stage, we might want to write a renovate package rule that moves/bumps `@prisma/*` and `nexus-plugin-prisma` together.

### DIFF
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -11,7 +11,12 @@
       "baseBranchList": ["latest"],
       "packageNames": ["nexus-plugin-prisma"],
       "enabled": true
+    },
+    {
+      "baseBranchList": ["latest"],
+      "packageNames": ["@prisma/cli", "@prisma/client"],
+      "enabled": true,
+      "updateTypes": ["major"]
     }
-  ],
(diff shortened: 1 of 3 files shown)


nikolasburk (approved):
lgtm 👍

PR #2209, opened 2020-11-10
fix: add build command to heroku example

2color:


divyenduz (commented):
Is this required or just better practices? Isn't implicit generate enough?

2color:
I just went through the Heroku guide and the implicit generate didn't work.

janpio:
Can you please open a bug report for that? It totally should, so this needs to be investigated.

2color:
Opened an issue: https://github.com/prisma/prisma/issues/4210

PR #2279, opened 2020-11-26
fix(testing-express): test fail on error

ryands17:


nikolasburk:
Thanks for the quick fix Ryan 🙌

PR #2288, opened 2020-11-30
refactor: 🚀 #1931 updated prisma schema 

### LINKED ISSUE #1931: Schema in the "prisma-examples/experimental/typescript/graphql" example seems to contain an error
I was trying to run the "prisma-examples/experimental/typescript/graphql" example and the schema is complaining about a reference that is wrong.

The definition for Post in the current schema is as follows:
``` prisma
model Post {
  id        Int     @id @default(autoincrement())
  title     String
  content   String?
  published Boolean @default(false)
  author    User?
}
```
Vscode indicates an error here on the author reference. 
I managed to fix it by replacing Post with the following:
``` prisma
model Post {
  id        Int     @id @default(autoincrement())
  title     String
  content   String?
  published Boolean @default(false)
  author    User? @relation(fields: [authorId], references:[id])
  authorId Int?
}
```

ruheni:
Updated Prisma schema in `experimental/typescript/graphql`

### DIFF
--- a/experimental/typescript/graphql/prisma/schema.prisma
+++ b/experimental/typescript/graphql/prisma/schema.prisma
@@ -1,18 +1,10 @@
-datasource db {
-  provider = "sqlite"
-  url      = "file:./dev.db"
-}
-
 generator client {
   provider = "prisma-client-js"
 }
 
-model Post {
-  id        Int     @id @default(autoincrement())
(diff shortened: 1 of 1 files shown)


nikolasburk:
Thanks Alex 🚀🚀🚀

PR #2310, opened 2020-12-04
refactor: updated nextjs examples 

### LINKED ISSUE #2303: Next.js examples should use only a single `PrismaClient` instance
Right now all Next.js examples instantiate `PrismaClient` multiple times. Instead, it should be instantiated only once and imported across the app like [here](https://github.com/prisma/blogr-nextjs-prisma/blob/final/lib/prisma.ts).

ruheni:
Updated nextjs examples to use a single `PrismaClient` instance #2303


nikolasburk:
🔥 🔥 🔥

PR #2313, opened 2020-12-05
fix: Missing Authorization Header error

### LINKED ISSUE #2295: Authorization Header not working
Hi,
I'm trying to get the typescript/graphql-auth example running, but getting stuck on authorising a request using the Authorization-Header. I get the graqphQL response `"Not Authorised!"` and my code seems to fail in function `getUserId` (utils.ts) at command ```const Authorization = context.request.get('Authorization')``` with the error `Cannot read property 'get' of undefined`.
In the `experimental/typescript/graphql-auth` example with graphql-yoga this seems to work. But in normal typescript/javascript example I have no success :-(

ruheni:
Fixed graphql-auth examples error #2295

### DIFF
--- a/javascript/graphql-auth/src/index.js
+++ b/javascript/graphql-auth/src/index.js
@@ -20,10 +20,10 @@ const server = new ApolloServer({
     }),
     permissions
   ),
-  context: (request) => {
+  context: ({ req }) => {
     return {
-      ...request,
-      prisma,
+      ...req,
+      prisma
     }
   },
 })
--- a/javascript/graphql-auth/src/utils.js
+++ b/javascript/graphql-auth/src/utils.js
@@ -2,8 +2,8 @@ const { verify } = require('jsonwebtoken')
 
 const APP_SECRET = 'appsecret321'
 
-function getUserId(context) {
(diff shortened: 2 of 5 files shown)


janpio:
Any idea why this did not get caught by the tests we were running on all these examples?

ruheni:
I believe the tests only check if there's a response from the `feed` query. This was more of an authentication issue on the protected mutations and queries that required a bearer token to be added in the authorization header.

nikolasburk:
> Any idea why this did not get caught by the tests we were running on all these examples?

Because these were runtime errors that only occur on specific GraphQL operations that require authentication and not on the generic, basic query we're testing.

janpio:
Would it make sense to add these as a test to these examples then?

nikolasburk:
> Would it make sense to add these as a test to these examples then?

Yes, this is already in the "examples maintenance" backlog [here](https://www.notion.so/prismaio/Example-Cleanup-Polish-0e3bd742d31343aca9cb575206928311#7057766994b44765b963464a7c509c2d). I definitely want to have project-specific tests eventually rather than just a generic query 👍

PR #2339, opened 2020-12-11
chore: clean up experimental

divyenduz:
Clean up after https://github.com/prisma/prisma-examples/pull/2327

### DIFF
--- a/.github/scripts/check-for-update.sh
+++ b/.github/scripts/check-for-update.sh
@@ -57,7 +57,7 @@ while [ $i -le $count ]; do
     echo "checking $item"
 
     case "$item" in
-      *".github"*|*"experimental"*)
+      (*".github"*)
         echo "ignoring $item"
         continue
         ;;
--- a/.github/scripts/check-for-update.sh
+++ b/.github/scripts/check-for-update.sh
@@ -57,7 +57,7 @@ while [ $i -le $count ]; do
     echo "checking $item"
 
     case "$item" in
-      (*".github"*)
+      *".github"*)
         echo "ignoring $item"
         continue
         ;;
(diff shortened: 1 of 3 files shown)


.github/scripts/check-for-update.sh
```diff
@@ -57,7 +57,7 @@ while [ $i -le $count ]; do
     echo "checking $item"
 
     case "$item" in
-      *".github"*|*"experimental"*)
+      (*".github"*)
```
janpio: Are you sure this `(` belongs here if it wasn't there before?

.github/scripts/upgrade-all.sh
```diff
@@ -14,7 +14,7 @@ dir=$(pwd)
 
 echo "$packages" | tr ' ' '\n' | while read -r item; do
   case "$item" in
-    *".github"*|*"experimental"*|*"deployment-platforms"*)
+    (*".github"*|*"deployment-platforms"*)
```
janpio: see https://github.com/prisma/prisma-examples/pull/2339/files#r540843088

.github/scripts/check-for-update.sh
divyenduz: Both are valid `sh` (checked with `shellcheck`), I think I should get rid of it though, looked up the case syntax and it doesn't need the first bracket.

divyenduz: https://devhints.io/bash (search for case/switch)

janpio: Weird syntax is weird 🤷

janpio:
Also requested review from @ruheni as he removed the experimental examples.

janpio (approved):

PR #2352, opened 2020-12-15
fix(rest-nextjs*): small improvements

EarlGeorge:
Deleting unused variables, adding type check and couple more improvements.


nikolasburk:
@EarlGeorge I'm really sorry for the delay on this one. And thank you so much for taking the time to improve this example, I think all improvements are really great! 🎉 🎉 🎉

typescript/rest-nextjs-express/frontend/package.json
```diff
@@ -18,7 +18,8 @@
     "react-markdown": "^5.0.0"
   },
   "devDependencies": {
-    "@types/node": "14.14.28",
+    "prisma": "2.17.0",
```
janpio: I think this dependency had been removed intentionally in this project.

nikolasburk: Ha, you're right! Thanks for catching this Jan!! 🦅 👁️  Fixing via: https://github.com/prisma/prisma-examples/pull/2577

PR #2408, opened 2021-01-11
fix: Profile schema in typescript/graphql example

ghost:
Schema currently has the field user reference itself in the relation, when it should be using the field userId. Fixed here.


nikolasburk:
Thanks for much for catching this @Zireael13 🚀 🚀 🚀

PR #2472, opened 2021-02-02
fix: update vercel deployment example

2color:


divyenduz:
@2color The tests don't seem to be running for this?

2color:
Not sure why, but I just rebased with latest and pushed. Seems like the tests are queued now.

divyenduz (approved):
✅

PR #2545, opened 2021-02-16
ci: Update setup-node usage

janpio:
To get rid of 
image

Also:
- better name for `debug` step
- better filename for sql server test


Jolg42 (approved):
Looks good

PR #2654, opened 2021-03-16
fix(deployment/heroku): Empty devDependencies to unblock scripts

janpio:


2color:
Moving prisma to dependencies was done because Heroku prunes dev dependencies, preventing you from running `migrate` commands in the release phase. 

I know this isn't idea, but this was a trade-off to ensure best practices.

janpio:
Ok, empty devDependencies should be enough to unblock the script - but the example might not be updated properly.

PR #2667, opened 2021-03-19
fix: Prisma schema and API update

ruheni:
Updated `schema.prisma` file and rest-express API example to match GraphQL examples


typescript/rest-express/src/index.ts
```diff
@@ -1,16 +1,25 @@
-import { PrismaClient } from '@prisma/client'
-import * as bodyParser from 'body-parser'
```
nikolasburk, line 2: nice! just a little reminder if we replace `bodyParser` here we can also remove the dep from `package.json`

typescript/rest-express/src/index.ts
```diff
@@ -48,6 +84,27 @@ app.delete(`/post/:id`, async (req, res) => {
...
+})
+
+app.get('/user/:id/drafts', async (req, res) => {
+  const { id } = req.params
+  const { email } = req.body
+
+  const drafts = await prisma.user.findUnique({
+    where: {
+      id: Number(id) || undefined,
+      email: email || undefined
```
nikolasburk: can you double check whether the `|| undefined` is actually needed here? I believe this was specific to Nexus so we _should_ be able to omit it in non-Nexus project. if it doesn't produce any compile-time or runtime errors, we can leave it out.

ruheni: This isn't needed. Thank you for spotting it. 🙂

typescript/rest-hapi/src/plugins/posts.ts
```diff
@@ -66,12 +66,30 @@ export default usersPlugin
...
+    const posts = await prisma.post.findMany({
+      where: {
+        published: true,
+        ...or
+      },
       include: { author: true },
+      take: Number(take) || undefined,
+      skip: Number(skip) || undefined,
+      orderBy: {
+        updatedAt: orderBy || undefined
```
nikolasburk, line 88: are these `|| undefined`s necessary here?

ruheni: Yes. If not added, it defaults to `NaN`.
Here's the error that is thrown if ` || undefined` is omitted: 
```bash
PrismaClientValidationError:
Invalid `prisma.post.findMany()` invocation:

{
  where: {
    published: true
  },
  include: {
    author: true
  },
  take: NaN,
        ~~~
  skip: NaN,
        ~~~
  orderBy: {
    updatedAt: 'desc'
  }
}
```

nikolasburk:
Great job Alex!!! 🚀

PR #2679, opened 2021-03-23
fix: removed seed script from script examples

ruheni:
- deleted seeding scripts
- updated script examples READMEs


nikolasburk:
Thank you Alex 🎉

PR #2695, opened 2021-03-27
fix: npm build failing

Primajin:
fix `Error can only be default-imported using the 'esModuleInterop' flag` during `npm run build`

image


ruheni:
Hi @Primajin 👋🏾

Thank you for fixing this! 🙏🏾

PR #2733, opened 2021-04-09
feat: Added Newman tests to scripts

ruheni:
- Updated GraphQL and REST test scripts to mirror README getting started instructions


.github/tests/javascript/graphql-sdl-first/run.sh
```diff
@@ -2,13 +2,14 @@
 
 set -eu
 
-yarn 
-yarn prisma db push --preview-feature
-yarn dev &
+npm install
+npx prisma migrate dev --name init
+npx prisma db seed --preview-feature
+npm run dev &
```
janpio, line 8: Hmm, no yarn lockfile in the repo that also needs upgrading? No other uses of `yarn` in script or similar? Then this might be fine although in Engineering we usually use yarn.

.github/tests/javascript/graphql-sdl-first/run.sh
```diff
@@ -2,13 +2,14 @@
...
 pid=$!
 
-sleep 10
+sleep 20
 
-curl --fail 'http://localhost:4000/' -H 'Accept-Encoding: gzip, deflate, br' -H 'Content-Type: application/json' -H 'Accept: application/json' -H 'Connection: keep-alive' -H 'DNT: 1' -H 'Origin: http://localhost:4000' --data-binary '{"query":"query {\n  feed {\n    id\n    content\n    author {\n      id\n      name\n      email\n    }\n  }\n}"}' --compressed
+npx newman run https://raw.githubusercontent.com/prisma/prisma-examples/latest/.github/tests/postman_collections/graphql.json --bail
```
janpio: Would it possible make sense to keep both the newman tests and the curl sanity check that was there before?

janpio: Can we not run this with a path instead of an absolute URL?

janpio (approved):
This looks great!

Added a few comments for details.

.github/tests/javascript/graphql-sdl-first/run.sh
ruheni: The curl sanity check wouldn't be as helpful when working with GraphQL as it always returns an HTTP 200, except when there's a server failure. 

The tests from the postman collection validate the status code and response body of every request.

.github/tests/javascript/graphql-sdl-first/run.sh
ruheni: `npm` is used to mirror the commands the user/learner would use when following the instructions on the README file of the example. `yarn` would work just fine as well.

.github/tests/javascript/graphql-sdl-first/run.sh
ruheni: I tried that; it didn't work.

Each test is only "aware" of the example and can't access any resource outside the example.

.github/tests/javascript/graphql-sdl-first/run.sh
janpio: Checked myself, we are not using lockfiles in this repo so this seems fine.

.github/tests/javascript/graphql-sdl-first/run.sh
janpio: That does not sound correct to me. You should be able to test the examples locally by running the local tests. Can we not just supply a relative path?

ruheni: I tried using a relative path in this [Action's run](https://github.com/prisma/prisma-examples/runs/2256064637?check_suite_focus=true#step:5:115), it couldn't find the file.

image

janpio: The script currently probably runs in the project folder context, so `/javascript/rest-express`. So paths would need to be relative to that or change that context first.

ruheni: That would lead to some redundancy, especially for the graphql examples. I opted for a central location for all the tests.

janpio: No, you can still leave the files where they are. But you reference them with a relative path to where the script is run (as the script is not run in the folder where it lives, but in the project folder).

Example could be: 
```suggestion
npx newman run ../../.github/tests/postman_collections/graphql.json --bail
```

ruheni: Sorry, I misunderstood that.
Let's give it a try..

ruheni: That worked! I'll update the other tests

PR #2769, opened 2021-04-19
feat(deployment-platforms/aws-lambda): add exclusion for unneeded binaries

janpio:
To avoid problems like https://github.com/prisma/prisma/issues/6644, we want to filter out all binaries but the required on during the packaging step of Serverless Framework.

Related PR in docs that highlights this: https://github.com/prisma/docs/pull/1578
Related PR adding this to e2e tests as well: https://github.com/prisma/e2e-tests/pull/1637

### DIFF
--- a/deployment-platforms/aws-lambda/serverless.yml
+++ b/deployment-platforms/aws-lambda/serverless.yml
@@ -38,3 +38,9 @@ functions:
       - http:
           method: GET
           path: /posts
+
+# only include the Prisma binary required on AWS Lambda while packaging
+package:
+  patterns:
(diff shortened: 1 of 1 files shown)


2color (approved):

janpio:
Note: We are currently investigating an issue where this rule is not enough and we need to understand if there is more we need/have to/can do - but I am merging this now anyway as the docs also refer to it.

2color:
> Note: We are currently investigating an issue where this rule is not enough and we need to understand if there is more we need/have to/can do - but I am merging this now anyway as the docs also refer to it.

Thanks. I'll keep an eye out for this.

PR #2934, opened 2021-06-12
fix: Sort contents of package.json 

### LINKED ISSUE #2907: Change order of dependencies in `latest`
Right now the dependencies in `latest` `package.json` files are not automatic but manual (`prisma` is often listed first). That leads to a more complicated than necessary diff in the autoamted branches like `dev` after we run `npm install prisma@dev` to get the latest `dev` version there, for example: https://github.com/prisma/prisma-examples/compare/latest...dev

We should clean up the order in `latest` once by running `npm install prisma @prisma/client` in all projects manually and commit that change.

ruheni:
Fix #2907 

Used `npx sort-package-json` to sort the order of dependencies and contents of `package.json` of all examples

### DIFF
--- a/databases/mongodb/package.json
+++ b/databases/mongodb/package.json
@@ -2,6 +2,9 @@
   "name": "prisma-mongo",
   "version": "1.0.0",
   "description": "",
+  "keywords": [],
+  "license": "ISC",
+  "author": "",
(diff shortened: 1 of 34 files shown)


janpio (approved):
<3

PR #2956, opened 2021-06-20
fix(javascript/rest-nuxtjs): wrong `signup` endpoint onSubmit

ktopouzi:
Hey there 👋 I recently tried to test the Nuxt - Prisma example and I noticed that the `signup` onSubmit action targeted the wrong endpoint (singup instead of user).

I decided to change the actual fetch instead of the endpoint but if you think it's the other way around please let me know. 

Hope this helps!

### DIFF
--- a/javascript/rest-nuxtjs/pages/signup.vue
+++ b/javascript/rest-nuxtjs/pages/signup.vue
@@ -27,7 +27,7 @@ export default {
           email: this.email,
         }
 
-        const res = await fetch(`http://localhost:3000/api/signup`, {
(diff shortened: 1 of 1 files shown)


ruheni:
Hey,

Thank you for spotting this. 🙏🏾

We'll also expand our tests to capture these minor errors. 🙂

ktopouzi:
No problem at all! It's kinda weird it's still failing for the semantic title 🤷‍♂️ If there's anything else I can help with, let me know 👍

janpio:
It is quite strict about that, so it required it to be lower cased. I updated it to also include the scope of the example project your change is about 👍

PR #2977, opened 2021-06-29
chore: Update README.md

pkellner:
This small bug (prisma is a global now and now using "context" which IMHO it really should, even if in context it refers to the global.)

At any rate, this fix is just for the graphql-nextjs example. I believe this error is propagated through all the other examples also.

(Also, a little unrelated and too big to fix for me, getInitialProps is outdated in this example)

### DIFF
--- a/typescript/graphql-nextjs/README.md
+++ b/typescript/graphql-nextjs/README.md
@@ -350,7 +350,7 @@ const Mutation = objectType({
 +       bio: stringArg()
 +     }, 
 +     resolve: async (_, args, context) => {
-+       return context.prisma.profile.create({
++       return prisma.profile.create({
 +         data: {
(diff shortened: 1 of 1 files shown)


janpio (changes requested):
Thanks for this change. The README is generated though, so someone needs to figure out the right location where the templates need to be adapted for this.

ruheni:
Hey @pkellner, 👋🏾 

I'd recommend taking a look at our [contributing guide](https://github.com/prisma/prisma-examples/blob/latest/CONTRIBUTING.md#improving-a-readme) on improving existing READMEs. 🙂

ruheni:
Closing this in favour of #3342

#3342 contains more updates on the template README.

PR #2986, opened 2021-07-02
fix(script.js): Add missing log output

janpio:
This is probably missing in other examples that do similar things.

### DIFF
--- a/javascript/script/script.js
+++ b/javascript/script/script.js
@@ -45,7 +45,7 @@ async function main() {
     },
   })
   console.log(
-    `Created users: ${user1.name} (${user1.posts.length} post) and (${user2.posts.length} posts) `,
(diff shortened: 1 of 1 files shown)


javascript/script/script.js
```diff
@@ -45,7 +45,7 @@ async function main() {
     },
   })
   console.log(
-    `Created users: ${user1.name} (${user1.posts.length} post) and (${user2.posts.length} posts) `,
+    `Created users: ${user1.name} (${user1.posts.length} post) and ${user2.name} (${user2.posts.length} posts) `,
```
2color:
```suggestion
    `Created users: ${user1.name} (${user1.posts.length} posts) and ${user2.name} (${user2.posts.length} posts) `,
```

PR #3020, opened 2021-07-13
fix: use replica set docker image in mongo example

2color:


janpio (commented):
Tests indicate this work, but further look shows that the tests are actually skipped:
```
running ./databases/mongodb/package.json
ignoring ./databases/mongodb/package.json
```

Change looks good anyway.

2color:
That's because of https://github.com/prisma/prisma-examples/blob/mongo-preview-fix/.github/scripts/test-all.sh#L18

I'd need to dive deeper into the test suite to make sure we're running the tests for this.

2color:
Tests fixed! 
Screen Shot 2021-07-14 at 16 09 32

janpio:
Nice work 💚

janpio:
Follow up from this: https://github.com/prisma/prisma-examples/issues/3022

Issue #1875, opened 2020-07-16
TypeScript + Next.js: Switch from getInitialProps to getStaticProps (SSG)


karlhorky:
[Next.js 9.3 introduced static-site generation](https://nextjs.org/blog/next-9-3#next-gen-static-site-generation-ssg-support), moving from `getInitialProps` to `getStaticProps` and `getServerSideProps`.

The `typescript/graphql-nextjs` example [uses `getInitialProps`](https://github.com/prisma/prisma-examples/blob/fea790dd311e532a6289c7de53f9dba05f36773b/typescript/graphql-nextjs/apollo/client.js#L41), which prevents use of these new APIs in pages.

Is it possible to switch to the new APIs? The following pull request in the Next.js examples did the same switch and can be used as a reference:

https://github.com/vercel/next.js/pull/13742

koniuszy:
@ruheni 
I believe we can close it now 👌 

ref: https://github.com/prisma/prisma-examples/pull/3033

karlhorky:
@koniuszy am I understanding correctly that #3033 no longer does any server-side / static build-time fetching (eg. which was done previously via `getInitialProps`)?

I guess if this is the case, it is a slight regression in that it is fetching the data only once React actually renders...?

If this is the case, for an example that uses Next.js to its full potential, my issue #1875 is still valid and should be reopened.

cc @ruheni

koniuszy:
@kandros right, 
since we had the loading condition working I believe it wasn't implemented correctly 🤷‍♂️ 

I've removed loading conditions and added the server fetching on build time,

the only issue I see is a failing build due to not hosted gql server - probably there is nothing we can do about it,
i've added a comment: https://github.com/prisma/prisma-examples/pull/3056/files#diff-1dc6e8f388abb6745bc3f068eb339d4ff2faab2225d5245671c99c37fdb219daR19

karlhorky:
@ruheni thanks for #4232 ! 🙌

I've opened #4234 to remove `legacyBehavior` and simplify the Next.js `<Link>` component usage.

PR #3033, opened 2021-07-26
chore: update typescript/graphql-nextjs example

### LINKED ISSUE #3041: missed a "seed.ts"  file
examples\graphql-nextjs\prisma  ,missed a "seed.ts"  file

koniuszy:
- upgrading nextjs
- using newest @apollo/client lib instead of: @apollo/react-hooks,  @apollo/react-ssr, apollo-boost, apollo-link-http, apollo-link-schema, react-apollo
- moving connectors from all pages to `_app` as 1 provider.
- upgrading apollo-server-micro to 3 (which introduced newest apollo-gql-playground). due to upgrade had to use `cors` and install `micro`
- introducing the nextjs way to expose handler on `/api/` route
- adding prettier for formatting consistency
- moving generated from `/api/` dir to prevent creating pages (nextjs routing)


ruheni:
Hmm, I was testing this out, and it seems to be missing the seed script. I'll go ahead and add it too for it to match the instructions on the README.

koniuszy:
> Hmm, I was testing this out, and it seems to be missing the seed script. I'll go ahead and add it too for it to match the instructions on the README.

@ruheni 
I fixed the missing seed.ts file and added adjustments around

ruheni:
Thanks for the update!

Could you update `.github/readmes/_using-the-graphql-api-nextjs.md]` to match the contents in this [commit](https://github.com/prisma/prisma-examples/commit/98ca9213f5e5399052c31d5748965fb965ebb2dd)? After that, change directory to [./github/tools](./github/tools) and run `npm run build-readmes`, commit and push the changes.

I've tested everything and after this update, we can finally merge the PR! 🍾

ruheni:
Thank you so much @koniuszy for the contribution! 💚

karlhorky:
@koniuszy am I understanding correctly that this no longer does any server-side fetching (eg. that was done previously via `getInitialProps`)?

I guess if this is the case, it is a slight regression in that it is fetching the data only once React actually renders...?

If this is the case, for an example that uses Next.js to its full potential, my issue #1875 is still valid and should be reopened.

cc @ruheni

ruheni:
@karlhorky, that makes sense, we should have SSR. I'll re-open the issue. 

@koniuszy you'd like to tackle?

PR #3054, opened 2021-08-08
fix(graphql-nextjs): update prisma client version

ruheni:


janpio:
Any idea why Renovate did not take care of this?

ruheni:
I merged a PR from a community member that was using an older version of Prisma which I missed and had to make the update

janpio:
A future Renovate run would also have fixed that again I think - but of course this was faster.

PR #3056, opened 2021-08-08
chore: extend typescript/graphql-nextjs example with permissions and SSG hydrating

### LINKED ISSUE #1604: Add user rules and permissions to typescript/graphql-nextjs
Please add example that shows how to check that a user has the permission to create/delete a post.

koniuszy:
Closes: #1604

- adding graphql shield as permissions manager
- adding global typed context
- new backend architecture due to overloading api/index.ts 
- resolvers sugar syntax: objectType => mutationField,  objectType => mutationField
- setting the apollo-server based on the official example: https://github.com/vercel/next.js/blob/canary/examples/api-routes-graphql/pages/api/graphql.js
- adding ssr fetching on build time and hydrating clients

@ruheni the last missing piece here is the codegen (react-hooks generator based on schema) - i don't want to overload the example with features but if u say a word - i will add it 🥲


ruheni (commented):
Hey @koniuszy 👋🏾 

Thank you for your contribution to this example. This is definitely an example that would be very helpful.

A few comments on this PR:
- Instead of extending the existing example, you can create a new example – `graphql-nextjs-auth`
- For the authorization, I think we can borrow from the existing [`graphql-auth`](https://github.com/prisma/prisma-examples/tree/latest/typescript/graphql-auth) for the permissions and rules
- For the type imports, we can make use of TypeScript's type-only imports, for example: `import type { NextApiRequest, NextApiResponse } from 'next'`
- Let's keep the example as simple as possible. If we can do away with lodash and deepmerge, that would be fantastic!
- We can move `types.ts` and `resolvers.ts` into `schema.ts`
- Let's keep the loading states in the example too

On the codegen, I think that would deserve a separate example - exploring an end-to-end type-safe stack (Apollo-codegen, GenQL, or GraphQL codegen). You can create an issue for this, and we can discuss what tools we could use to accomplish it. 

_Could be time to explore another GraphQL client for our examples other than Apollo Client to simplify our examples - deserves another issue._ 🤔 

I'm sorry for the late response, and I hope you will find this feedback useful.

ruheni:
Hi @koniuszy, will you continue working on this PR?

PR #3079, opened 2021-08-13
test: Add e2e tests to rest-nextjs-api-routes example

pshaddel:
Implementing e2e tests using Playwright


pshaddel:
Hi @ruheni 
I started working on this and that would be awesome if you take a look at it.
* Also I am not able to run this test inside `.github` folder. It is working when I run this file in other places:
```bash
npx playwright test ./test.spec.ts
```
Do you have any idea why this is happening?

* I added `id` to some inputs to select them using playwright.
* I saw that if we register an email twice it returns 500 error but it redirects user to blog page like nothing happened I think we can fix it in another issue.
* Right now I'm counting on seed data. I'm not sure if this is a good idea.

pshaddel:
@ruheni Hey. Did you have time to take a look at this?

jdmachogg:
Hey @pshaddel - to comply with US Sanctions administered by OFAC, we unfortunately cannot accept a PR from anyone located in Iran. We regret this, and hope that there is a solution in the future. Thank you and all the best from us at Prisma.

PR #3091, opened 2021-08-28
fix: misspelling of secret env variable

shimamooo:
Expected RANDOM_STRING, instead there is RAMDOM_STRING


ruheni:
Thanks for catching this @AnthonyKuang!  🙌🏾

PR #3123, opened 2021-09-01
chore: update apollo server example

m-abdelwahab:
This PR updates both JS and TS GraphQL examples that were using Apollo Server 2.

Affected examples:
- JavaScript
  -  `graphql`
  - `graphql-auth`
  - `graphql-sdl-first`
- TypeScript
  - `graphql`
  - `graphql-auth`
  - `graphql-sdl-first`
  - `graphql-typegraphql` 
  - `graphql-typegraphql-crud`

### DIFF
--- a/typescript/graphql-sdl-first/src/schema.ts
+++ b/typescript/graphql-sdl-first/src/schema.ts
@@ -1,8 +1,8 @@
-import { makeExecutableSchema } from 'apollo-server'
+import { gql } from 'apollo-server'
 import { DateTimeResolver } from 'graphql-scalars'
 import { Context } from './context'
 
-const typeDefs = `
+export const typeDefs = gql`
 type Mutation {
   createDraft(authorEmail: String!, data: PostCreateInput!): Post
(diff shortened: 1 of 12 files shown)


ruheni:
LGTM! 🚀

Thanks for taking care of this.

Sidenote: we should probably consider adding a prettier script to format all examples and create some form of consistency in formatting. Probably worth a separate issue we should create and tackle it later.

typescript/graphql-sdl-first/src/schema.ts
```diff
@@ -214,8 +214,3 @@ interface UserCreateInput {
   posts?: PostCreateInput[],
 }
```
ruheni, line 216: Is there any reason why we're not using `makeExecutableSchema`?

m-abdelwahab: image
image
https://github.com/apollographql/apollo-server/blob/main/CHANGELOG.md

It got removed in this version

PR #3174, opened 2021-09-21
Change NestJS example

gustavonobreza:
In Nestjs documentation, the example with Prisma is like this.


ruheni:
Hi @gustavonobreza 👋🏽

Thank you for bringing this up. 

I'll close this PR in favor of #3344 because this PR also affects a lot of examples. The Prisma versions in the affected examples are also out of date. 🙂

Issue #3153, opened 2021-09-13
Seed command fails


hnrw:
Running `npx prisma db seed`  in the `rest-nextjs-api-routes` gives the following error.

```
An error occured while running the seed command:
/home/henry/projects/prisma/rest-nextjs-api-routes/prisma/seed.ts:37
import { PrismaClient } from '@prisma/client';
^^^^^^

SyntaxError: Cannot use import statement outside a module
    at wrapSafe (internal/modules/cjs/loader.js:1117:16)
    at Module._compile (internal/modules/cjs/loader.js:1165:27)
    at Module.m._compile (/home/henry/projects/prisma/rest-nextjs-api-routes/node_modules/ts-node/src/index.ts:1310:23)
    at Module._extensions..js (internal/modules/cjs/loader.js:1221:10)
    at Object.require.extensions.<computed> [as .ts] (/home/henry/projects/prisma/rest-nextjs-api-routes/node_modules/ts-node/src/index.ts:1313:12)
    at Module.load (internal/modules/cjs/loader.js:1050:32)
    at Function.Module._load (internal/modules/cjs/loader.js:938:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:71:12)
    at main (/home/henry/projects/prisma/rest-nextjs-api-routes/node_modules/ts-node/src/bin.ts:331:12)
    at Object.<anonymous> (/home/henry/projects/prisma/rest-nextjs-api-routes/node_modules/ts-node/src/bin.ts:482:3)
```

I followed the steps given in the readme: https://github.com/prisma/prisma-examples/tree/latest/typescript/rest-nextjs-api-routes

Interestingly, seeding the database works fine in the typescript rest-express project. 

I'm not too experienced with TS, so maybe there's something obvious I'm missing. But it would be better if the example worked out of the box.

Ningensei848:
In my case, I solved the problem by running the `seed` command with the compiler option. (Rewrite your `package.json` like below)

cf. https://github.com/prisma/prisma-examples/issues/2668#issuecomment-803274884

```package.json
.
.
    "prisma": "3.0.2",
    "typescript": "4.4.3"
  },
  "prisma": {
    "seed": "ts-node --compiler-options {\"module\":\"CommonJS\"} prisma/seed.ts"
  }
```

---

in terminal:

```terminal
$ npx prisma db seed
Environment variables loaded from .env
[...]

PR #3192, opened 2021-09-26
Fix issue #3153

tromgy:
`ts-node` needs to run with `{"module": "commonjs"}` compiler option, otherwise the `npx prisma db seed` command fails.

### DIFF
--- a/typescript/graphql-nextjs/package.json
+++ b/typescript/graphql-nextjs/package.json
@@ -40,6 +40,6 @@
     "typescript": "4.4.3"
   },
   "prisma": {
-    "seed": "ts-node prisma/seed.ts"
+    "seed": "npm run ts-node prisma/seed.ts"
   }
 }


janpio:
How is the PR description related to the file change?

tromgy:
Ah, I see. It's the same issue, just in a different example. I've run into it in the **graphql-nextjs** example, which this fixes. I can create a separate issue if you'd like. And maybe make a similar PR for **rest-nextjs-api-routes**

janpio:
Ah, I understand the PR now: Adding `npm run` makes it use the `ts-node` entry that exists further up in the `package.json` file.

In theory that should not be necessary, but if it is, it can directly be added to the `prisma.seed` entry insteaf of going the indirection of the `ts-node` entry.

tromgy:
I did it purposedly that way so that any necessary compiler flags for `ts-node` need to be specified only once.

ruheni:
Thanks for your PR to fix this @tromgy! But I'll go ahead and close it because #3254 fixes seeding for Next.js examples

PR #3311, opened 2021-11-08
fix(resolver/user): allUsers should return an array

jprando:
https://github.com/prisma/prisma-examples/blob/ea1048078c3502cedd5ccae53cf8991bf5fc767e/typescript/graphql-nestjs/src/resolvers.user.ts#L75

method allUsers informs that it returns a simple User, it should return an array of users.


ruheni:
Hey @jprando 👋🏽

Thank you for spotting this. 💚

The last bit is running the app for the `schema.gql` to be regenerated to update the query's return type. I'll take care of that and merge this in.

PR #3323, opened 2021-11-15
feat: delegated types with Prisma

ruheni:
Add union/ delegate type example


data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,64 @@
+# Data modeling: Delegate types
+
+This example shows you how to model [delegate types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a **Node.js script** to read and write data.
```
janpio: Should this maybe mention TypeScript?

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,64 @@
...
+
+This example consists of a `script.ts` file that contains example Prisma Client queries for delegate types
+
+
+## What are Delegate types?
+
+TODO?
```
janpio: TODO

janpio:
Are we planning to add test that make sure this example keeps working?
Generally we separate the examples by language used until now. Are we sure we want to change this for this example?

data-modeling/delegate-types/README.md
ruheni: I was conflicted on whether I should add an explanation on what delegate/ union types are in this section, or we could post something on dev.to

data-modeling/delegate-types/README.md
ruheni: Yes. I'll add that

ruheni:
@nikolasburk suggested we could add the new section for data modeling. This could expand too while we explore other patterns like Polymorphic Associations.

I'll add tests.

nikolasburk:
> Generally we separate the examples by language used until now. Are we sure we want to change this for this example?

That's a fair point @janpio, but I think for the data modeling example the focus is on the Prisma schema and not so much on the language itself so I think it's reasonable to highlight these examples only in one language.

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
+# Data modeling: Delegate types
+
+This example shows you how to model [delegate types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a **TypeScript script** to read and write data.
```
nikolasburk: Since the language is not too relevant for the data modeling example I'd de-emphasize it at this point

```suggestion
This example shows you how to model [delegate types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a TypeScript script to read and write data.
```

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+This example consists of a `script.ts` file that contains example Prisma Client queries for delegate types
```
nikolasburk: Isn't this sentence pretty much redundant with the previous paragraph?

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+
+## What are Delegate types?
```
nikolasburk:
```suggestion
## What are delegated types?
```

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+An application such as LinkedIn contains a feed that would contain different types of content such as articles, images, and videos. In such an instance, you would have a *parent entity*, `Feed`, containing all the shared properties and *child entities*,`Article`, `Image`, and `Video`, that inherit the *parent entity*'s properties and have their individual properties.
```
nikolasburk: I don't agree that the parent entity would be called `Feed` in this case since the feed itself is a different concept than articles, images or videos — it's basically a _sorted collection_ of these. I think `FeedItem` would be more appropriate.

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+This example shows you how to model [delegate types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a **TypeScript script** to read and write data.
+
+This example consists of a `script.ts` file that contains example Prisma Client queries for delegate types
+
+
+## What are Delegate types?
+
+An application such as LinkedIn contains a feed that would contain different types of content such as articles, images, and videos. In such an instance, you would have a *parent entity*, `Feed`, containing all the shared properties and *child entities*,`Article`, `Image`, and `Video`, that inherit the *parent entity*'s properties and have their individual properties. 
+
+When querying feed data, you would only query the feed data instead of querying the `Article`, `Image`, and `Video` tables and then merging them into one response. Querying and merging responses from the individual tables wouldn't allow you to perform operations such as ordering and filtering. This is the problem that delegate type solves.
```
nikolasburk: > When querying feed data, you would only query the feed data

Can you rephrase this to make this more clear?

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+Download this example:
+
+```
+curl https://codeload.github.com/prisma/prisma-examples/tar.gz/latest | tar -xz --strip=2 prisma-examples-latest/data-modeling/delegate-types
+```
+
+Install npm dependencies:
+```
+cd delegate-types
```
nikolasburk: We might want to rename the directory to "delegated-types"

data-modeling/delegate-types/README.md
```diff
@@ -0,0 +1,68 @@
...
+
+```
+cd prisma-examples/data-modeling/delegate-types
+npm install
+```
+</details>
+
+### 2. Create and seed the database
+
+Run the following command to create your SQLite database file. This also creates the `Content`, `Video`, `Image` and `Article` tables that are defined in [`prisma/schema.prisma`](./prisma/schema.prisma):
```
nikolasburk: Is `Content` equivalent to what you described as `Feed` (and I rather see as `FeedItem`) above?

nikolasburk (changes requested):

data-modeling/delegate-types/README.md
ruheni: Yes. I'll rename the references.

data-modeling/delegate-types/README.md
ruheni: > When querying the feed, you would only query the `Feed` table instead of querying ...

Is this clearer?

data-modeling/delegate-types/README.md
ruheni: I think this proves naming is indeed difficult 😆 

We could go with `Activity` perhaps. It was used in the [union types issue](https://github.com/prisma/prisma/issues/2505).

data-modeling/delegated-types/README.md
```diff
@@ -0,0 +1,67 @@
+# Data modeling: Delegate types
+
+This example shows you how to model [delegate types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a TypeScript script to read and write data.
```
nikolasburk:
```suggestion
This example shows you how to model [delegated types](https://edgeapi.rubyonrails.org/classes/ActiveRecord/DelegatedType.html) using Prisma, and use [Prisma Client](https://www.prisma.io/client) in a TypeScript script to read and write data.
```

data-modeling/delegated-types/README.md
```diff
@@ -0,0 +1,67 @@
+# Data modeling: Delegate types
```
nikolasburk:
```suggestion
# Delegated types
```

data-modeling/delegated-types/README.md
```diff
@@ -0,0 +1,67 @@
...
+When querying the feed, you would only query the Feed table instead of querying the `Article`, `Image`, and `Video` tables and then merging them into one response. Querying and merging responses from the individual tables wouldn't allow you to perform operations such as ordering and filtering. This is the problem that delegate type solves.
+
+The `Activity` response mimics union types by using delegate types.
+
+## How to use
+
+Download this example:
+
+```
+curl https://codeload.github.com/prisma/prisma-examples/tar.gz/latest | tar -xz --strip=2 prisma-examples-latest/data-modeling/delegate-types
```
nikolasburk: The folder is called `delegated-types`, right?

```suggestion
curl https://codeload.github.com/prisma/prisma-examples/tar.gz/latest | tar -xz --strip=2 prisma-examples-latest/data-modeling/delegated-types
```

data-modeling/delegated-types/README.md
```diff
@@ -0,0 +1,67 @@
...
+```
+
+Install npm dependencies:
+```
+cd delegate-types
```
nikolasburk:
```suggestion
cd delegated-types
```

data-modeling/delegated-types/README.md
```diff
@@ -0,0 +1,67 @@
...
+Clone this repository:
+
+```
+git clone git@github.com:prisma/prisma-examples.git --depth=1
+```
+
+Install npm dependencies:
+
+```
+cd prisma-examples/data-modeling/delegate-types
```
nikolasburk:
```suggestion
cd prisma-examples/data-modeling/delegated-types
```

PR #3459, opened 2022-01-10
docs: fix single tic

donny-son:
Modified unpaired tic in postgis-express documentation

### DIFF
--- a/typescript/postgis-express/README.md
+++ b/typescript/postgis-express/README.md
@@ -70,7 +70,7 @@ Execute this command to start the server:
 npm run dev
 ```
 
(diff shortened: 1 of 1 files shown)


janpio:
There might be a README template where this is taken from that also needs to be fixed.

donny-son:
don't think it's in this repo though

janpio:
It actually is: https://github.com/prisma/prisma-examples/blob/latest/.github/readmes/typescript/postgis-express/README.md

donny-son:
thanks! I'll close and resubmit a PR

PR #3479, opened 2022-01-15
fix: React Markdown syntax update (partial)

joseppu:
Post.tsx component and [id].tsx page in most (if not all) repos require this fix.

"source" is obsolete and should be replaced with "children". 

I am a newbie so changed just a single repo to be safe.


ruheni:
Thank you for pointing this out. 🙂

I created a PR that fixes all instances `react-markdown` is used in the repository. I'll therefore go ahead and close it in favor of #3511.

PR #3521, opened 2022-02-08
fix: (CockroachDB) remove instructions to add CA certificate

2color:
Since Cockroach uses Let's Encrypt, this shouldn't be necessary

### DIFF
--- a/databases/cockroachdb/README.md
+++ b/databases/cockroachdb/README.md
@@ -19,7 +19,7 @@ The example consists of two parts:
 - Node.js installed.
 - [Docker](https://www.docker.com/products/docker-desktop) installed
 
-> **Note:** You can also connect to a [free CockroachDB Serverless Cluster](https://www.cockroachlabs.com/docs/cockroachcloud/create-a-serverless-cluster.html). This requires [setting up a root certificate](https://www.cockroachlabs.com/docs/cockroachcloud/connect-to-a-serverless-cluster.html#step-2-connect-to-your-cluster).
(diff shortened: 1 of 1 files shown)


databases/cockroachdb/README.md
```diff
@@ -72,10 +72,7 @@ docker compose up -d
 
 Follow the following [guide](https://www.cockroachlabs.com/docs/cockroachcloud/create-a-serverless-cluster.html) to create a free CockroachDB Serverless cluster.
 
-After creating the database, you will need to:
-
-- [Download and configure the CA certificate](https://www.cockroachlabs.com/docs/cockroachcloud/connect-to-a-serverless-cluster.html#step-2-connect-to-your-cluster).
-- [Install the `cockroach` CLI](https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html) to use the the CockroachDB client (interactive shell) to run the `dbinit.sql` SQL script.
+After creating the database, you will need to [Install the `cockroach` CLI](https://www.cockroachlabs.com/docs/stable/install-cockroachdb.html) to use the the CockroachDB client (interactive shell) to run the `dbinit.sql` SQL script.
```
janpio, line 75: This can technically be done with any database UI or CLI that support PostgreSQL.

janpio (approved):

databases/cockroachdb/README.md
2color: Right! I give the exact command in a later step which is why I recommend installing the cockroach CLI.

PR #3667, opened 2022-04-10
refactor: improve TypeScript typing

altano:
The `typescript/graphql-nextjs` example was very helpful in getting me rolling. The one thing it was missing was good TypeScript (TS) typing. TS is enabled in the project but `strict` is set to `false` in tsconfig.json, and there is a lot of `any` types in the code. To improve this, I:

- Switched to `strict: true` in tsconfig.json
- Added @graphql-codegen packages to the project to automatically generate TS types from the GraphQL queries.
- Update `generate` package.json script to run this new codegen.
- Fix all the TS errors (e.g. implicit any types)

This should make this boilerplate much more strictly typed and generally useful.


typescript/graphql-nextjs/pages/index.tsx
```diff
@@ -53,11 +61,14 @@ const Blog = () => {
...
-            </div>
-          ))}
+          {data?.feed?.map(
+            post =>
+              post && (
+                <div key={post.id} className="post">
+                  <Post post={post} />
+                </div>
+              )
+          )}
```
ruheni, line 71: Let's conditionally render the data here. This should not be nullable because the query will always return either an empty array or an array that's populated.

typescript/graphql-nextjs/pages/drafts.tsx
```diff
@@ -53,11 +61,14 @@ const Drafts = () => {
       <div className="page">
         <h1>Drafts</h1>
         <main>
-          {data.drafts.map(post => (
-            <div key={post.id} className="post">
-              <Post post={post} />
-            </div>
-          ))}
+          {data?.drafts?.map(
```
ruheni, line 64: This shouldn't be nullable here. Let's use conditional rendering

typescript/graphql-nextjs/pages/drafts.tsx
```diff
@@ -18,28 +19,35 @@ const DraftsQuery = gql`
...
-          color: inherit;
-          padding: 2rem;
-          display: block;
-        }
-      `}</style>
-    </a>
-  </Link>
-)
+type Drafts = DraftsQuery["drafts"]
+type Draft = NonNullable<Drafts>[number]
```
ruheni, line 23: The generated GraphQL types have the correct types you can use when making queries and mutations, let's use that here.

Refer to [Apollo Client 3 with GraphQL Code Generator](https://youtu.be/PYDGjTufGsk) video by Jamie Barton or the repository he created https://github.com/notrab/graphql.wtf-graphql-codegen-apollo-client on how to use the generated types in the application.

ruheni: The same applies to the sections where types for queries have been manually defined - let's use the generated types.

ruheni (changes requested):
Hi @altano 👋🏽

Thanks for the contribution. This looks good. 

Before we merge this PR, could you:
- [ ] Exclude the generated files from the example
- [ ] Update the `.gitignore` file to exclude generated files
- [ ] Update the test for this example to include the `generate` script

jharrell:
Unfortunately, going to close this as abandoned. If you'd like, please open a new PR and we'll give it another look

PR #3795, opened 2022-05-25
Fix #3794

### LINKED ISSUE #3794: Unexpected token < in JSON at position 0
Hello,

I've found an issue with [this example](https://github.com/prisma/prisma-examples/blob/latest/typescript/graphql-nextjs/pages/api/index.ts), when running in production mode, sometimes the page would render:
Screenshot 2022-05-25 at 13 42 36

The issue occurred while running 3 parallel tests using Playwright on different browsers. I suspect that the ` schema = makeSchema({ ... })` would take some time to generate the files and it would crash.

robikovacs:


ruheni:
Thanks for catching this and your contribution. 🙏🏽

PR #3926, opened 2022-07-12
Update seed.ts

tapsu01:
Fix `email' field not unique


ruheni:
Hey 👋🏽

Could you clarify what error you ran into and how the suggested change resolves the error?

PR #3968, opened 2022-07-29
don't fetch from the api using a hardcoded host

fantapop:
In some cases, localhost:3000 won't be where the api is deployed to. For frontend code, just pass the path to fetch so it uses the same host.
Restructured server-side code to use library calls and avoid fetch which is considered a best practice for Next.js.

### DIFF
--- a/.github/tests/javascript/rest-nextjs/run.sh
+++ b/.github/tests/javascript/rest-nextjs/run.sh
@@ -9,8 +9,6 @@ pid=$!
 
 sleep 15
 
-curl --fail 'http://localhost:3000/api/feed?searchString=Prisma'
-
 # check frontend
 curl --fail 'http://localhost:3000/'
 
--- a/javascript/rest-nextjs/lib/util.js
+++ b/javascript/rest-nextjs/lib/util.js
@@ -1,0 +1,8 @@
+
+// Make an object serializable to JSON.
+//
+// Useful to convert an object which may contain non-serializeable data such as
+// Dates to an object that doesn't
+export function makeSerializable (o) {
+    return JSON.parse(JSON.stringify(o))
+}
--- a/javascript/rest-nextjs/pages/api/drafts.js
+++ b/javascript/rest-nextjs/pages/api/drafts.js
@@ -1,9 +1,0 @@
-import prisma from '../../lib/prisma'
-
-export default async function handle(req, res) {
-  const posts = await prisma.post.findMany({
-    where: { published: false },
-    include: { author: true },
-  })
-  res.json(posts)
-}
--- a/javascript/rest-nextjs/pages/api/feed.js
+++ b/javascript/rest-nextjs/pages/api/feed.js
@@ -1,9 +1,0 @@
-import prisma from '../../lib/prisma'
-
-export default async function handle(req, res) {
-  const posts = await prisma.post.findMany({
-    where: { published: true },
-    include: { author: true },
-  })
-  res.json(posts)
-}
--- a/javascript/rest-nextjs/pages/api/post/[id].js
+++ b/javascript/rest-nextjs/pages/api/post/[id].js
@@ -3,24 +3,13 @@ import prisma from '../../../lib/prisma'
 export default async function handle(req, res) {
   const postId = req.query.id
 
-  if (req.method === 'GET') {
-    handleGET(postId, res)
-  } else if (req.method === 'DELETE') {
+  if (req.method === 'DELETE') {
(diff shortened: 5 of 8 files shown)


ruheni:
Hey 👋🏽

Thank you for your contribution to the examples! Here are a few things to note:
- I would recommend keeping the API routes as they were
- I would also recommend using [`next-absolute-url` ](https://www.npmjs.com/package/next-absolute-url) to retrieve the origin from the request. It's a fairly small library and keeps things simple.

Looking forward to merging the PR! 🎉

fantapop:
Hey, Thanks for looking at this @ruheni .  I appreciate your dedication for keeping the examples simple and clear.  

Unfortunately, The next-absolute-url library won't help in this case because it requires prior knowledge of what the server host is.  When deploying to Vercel, edge functions are spun up to host the backend. The api and frontend are then hosted from a domain that is unique to that deployment. The code is not running on localhost:3000 and there is no provided mechanism to determine the host on the server.  

The changes in this PR implement the best practice described by nextjs here: https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props#getserversideprops-or-api-routes  but I can also appreciate if you'd rather keep your examples minimal.

Based on my testing, the two changes that would be required to allow these prisma example to be one click deployed to vercel are:

- this change (namely, don't hardcode server host for backend calls)
- provide the ability to use a remote database (sqlite doesn't work on vercel's read only file system)

We're already keeping a fork of prisma-examples which updates the db provider to be cockroachdb. If you decide you don't want the changes, we can keep this there instead.

ruheni:
Ah, I see. Thank you for the clarification. 

The `rest-nextjs-api-routes` example showcased how one could use Prisma in Next.js using API routes. I think we can simplify the example further by moving Prisma queries to `getServerSideProps` and doing away with the `GET API routes. The example contains other API routes, so I think this change would be reasonable.

fantapop:
Another idea that could solve this but help to keep things simple is to do away with the SSR in the examples.  I don't think SSR is key to showcasing Prisma's value and can add complexity in some cases.  In this case it allows using fetch for each call without passing a domain.  I could revert all the lib changes and instead update the few calls that are using getServerSideProps.  For example:

```diff
index 8011aad16..11f650824 100644
--- a/javascript/rest-nextjs/pages/index.jsx
+++ b/javascript/rest-nextjs/pages/index.jsx
@@ -1,15 +1,33 @@
 import Layout from '../components/Layout'
 import Post from '../components/Post'
-import { getFeed } from './api/feed'
-import { makeSerializable } from '../lib/util';
+import { useEffect, useState } from 'react';
+
+const Blog = () => {
+
+  const [feed, setFeed] = useState();
+
+  useEffect(() => {
+    const fetchData = async () => {
+      try {
+          const res = await fetch('/api/feed')
+          setFeed(await res.json())
+      } catch (error) {
+          console.log("error", error)
+      }
+    };
+    fetchData()
+  }, []);
+
+  if (!feed) {
+    return null;
+  }

-const Blog = props => {
   return (
     <Layout>
       <div className="page">
         <h1>My Blog</h1>
         <main>
-          {props.feed.map(post => (
+          {feed.map(post => (
             <div key={post.id} className="post">
               <Post post={post} />
             </div>
@@ -34,11 +52,4 @@ const Blog = props => {
   )
 }

-export const getServerSideProps = async () => {
-  const feed = await getFeed();
-  return {
-    props: { feed: makeSerializable(feed) },
-  }
-}
-

to further simplify the example you might be able to extract the useEffect out to a custom hook.
```

ruheni:
Hey,
I had some time to think over the example, and I think it would be a good idea to follow the Next.js best practices — making the Prisma query in `getServerSideProps`. The example can showcase both patterns — using `getServerSideProps` and API routes.

We can rename the examples to `rest-nextjs` (after merging this PR). I will take care of that.

fantapop:
Okay, sounds good @ruheni.  Are there still specific changes you want me to make to this PR before merging?

ruheni:
Yes, let's do away with the `GET` API Routes and move the Prisma queries to `getServerSideProps`

fantapop:
Hey @ruheni I've force pushed the new changes.  Is this inline what what you were suggesting?

ruheni:
Hey,
This looks good. The last minor change will be updating the `.github/tests/javascript/rest-nextjs/run.sh` by removing the `curl` request to `/api/feed` to get the tests passing.

fantapop:
made that change.  Hopefully tests passing now 🤞

PR #4061, opened 2022-09-06
chore: remove unsupported endpoint

marcoieni:
`GET /api/post/:id` is not implemented, so I removed it from the README

## How to reproduce

Follow the instructions to run the example and then run: `curl 127.0.0.1:3000/api/post/1`.

## Alternative solution for this issue

If you want to implement this GET endpoint, I can open an issue.

### DIFF
--- a/.github/readmes/_using-the-rest-api-nextjs.md
+++ b/.github/readmes/_using-the-rest-api-nextjs.md
@@ -4,7 +4,6 @@ You can also access the REST API of the API server directly. It is running on th
 
 ### `GET`
 
-- `/api/post/:id`: Fetch a single post by its `id`
 - `/api/feed`: Fetch all _published_ posts
 - `/api/filterPosts?searchString={searchString}`: Filter posts by `title` or `content`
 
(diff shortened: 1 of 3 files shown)


ruheni:
Hey @MarcoIeni 

Thank you for your contribution to the Prisma Examples. 

We use templates to auto-generate our READMEs located in [`./.github/readmes`](https://github.com/prisma/prisma-examples/blob/latest/.github/readmes) Update the relevant template and then navigate to `.github/tools`, run `npm install` and `npm run build-readmes` to generate one.

marcoieni:
I see, cool!
Maybe you can add a check in CI which detects if readmes diverge from the template.

ruheni:
Sure! Thanks for the idea. Do you have a GH Action in mind that we could user that handles that or would we have to build one from scratch? 🤔

In the mean time, I'll merge the PR. Thanks again for spotting this. 🙂

marcoieni:
> Do you have a GH Action in mind that we could user that handles that or would we have to build one from scratch? 🤔

I don't know any GH action that does this.

Thanks for merging the PR. Have a nice weekend!

PR #4147, opened 2022-10-07
adding declare to remove TS warning for "global.prisma"

umbrashia:
"global.prisma" is giving warning (Element implicitly has an 'any' type because type 'typeof globalThis' has no index signature.) because  is not define in global variable  and typescript is type safe, 

So that I am adding declare key word to define type safety "global.prisma" as per typescript standards


ruheni:
Thanks for the contribution @umbrashia. This will be really helpful. 🙏🏽

PR #4178, opened 2022-10-21
Update prisma.ts

thilllon:
declare clause with `var` causes redundant eslint-disabling code; `eslint-disable-next-line no-var`. I think there is no reason to use `var` in this case.


ruheni:
Hi @thilllon 👋🏽

Thanks for the contribution. I'll go ahead and merge this in! 🙏🏽

PR #4194, opened 2022-10-25
chore: skip yarn install step && make tests mirror steps in the READMEs

ruheni:


.github/scripts/test-all.sh
```diff
@@ -24,8 +24,8 @@ esac
 cd "$(dirname "$item")/"
 
 ## ACTION
-yarn install
-yarn prisma generate
+# yarn install
+# yarn prisma generate
```
ruheni:
```suggestion
```

PR #4234, opened 2022-11-07
Simplify Next.js Link component usage

karlhorky:
`legacyBehavior` prop added in #4232

### DIFF
--- a/typescript/graphql-nextjs/components/Header.js
+++ b/typescript/graphql-nextjs/components/Header.js
@@ -11,21 +11,19 @@ const Header = () => {
   return (
     <nav>
       <div className="left">
-        <Link href="/" legacyBehavior>
-          <a className="bold" data-active={isActive("/")}>
-            Blog
-          </a>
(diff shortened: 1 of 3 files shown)


ruheni:
Hey @karlhorky 👋🏽

Thanks for the contribution. I just tried out the changes you made and it seems to break the example's styles (see attached screenshot). 

image

karlhorky:
What should this look like? Maybe it's because `styled-jsx` is used here?

karlhorky:
I will switch from `styled-jsx` to inline styles :)

karlhorky:
@ruheni instead of switching to inline styles, I moved the style block to a different `<style>` tag. I guess this should be equivalent now.

ruheni:
The links should look something like this: 
image

The latest changes still doesn't resolve the styling

karlhorky:
@ruheni I haven't cloned the repo onto my machine, maybe you can try debugging why it's not working? I would recommend against using the `legacyBehavior` option - this will probably break in the future.

PR #4245, opened 2022-11-10
Set --transpile-only for Typescript

denizt:
When calling 'seed', if we don't have transpile-only flag set, we will end up with an empty database.

Set the flag so the example works as intended.


ruheni:
Thanks for the contribution 🙏🏽

PR #4310, opened 2022-12-08
Add pothos example

hayes:
Hey, wanted to add a Pothos example to this repo.

I'd also love to update a few of the other examples (or maybe some of the other Prisma/GraphQL documentation) to use pothos as well, but wanted to just start with a smaller change here.


typescript/graphql-pothos/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "typescript-graphql",
```
janpio:
```suggestion
  "name": "typescript-graphql-pothos",
```

typescript/graphql-pothos/src/server.ts
```diff
@@ -0,0 +1,22 @@
...
+    }
+  },
+})
+
+const server = createServer(yoga)
+
+server.listen(4000, () => {
+  console.log(`\
+🚀 Server ready at: http://127.0.0.1:4000
+⭐️ See sample queries: http://pris.ly/e/ts/graphql-pothos#using-the-graphql-api
```
janpio: This will need a follow up PR at https://github.com/prisma/pris.ly to work.

nikolasburk:
Hey Michael, this is amazing, thanks so much for contributing it!

FWIW, we just talked about this yesterday and we'd like to make Pothos the _default_ for our GraphQL examples indeed! So, if you like you can:
- rename the current `graphql` example to `graphql-nexus`
- rename your contributed example from `graphql-pothos` to just `graphql`

janpio:
Note this seems to be missing a test setup for now: https://github.com/prisma/prisma-examples/actions/runs/3645451073/jobs/6158548158#step:5:20

typescript/graphql-pothos/package.json
sre-is-toby: I think there's just a typo in the suggestion - did you mean `typescript-graphql-pothos`?

janpio: Something felt off when I type that, but I could not figure it out. Thanks for tell me - fixed above in the suggestion. Now it makes sense.

typescript/graphql-pothos/src/server.ts
hayes: I guess this doesn't apply to this one anymore, but the nexus one should be updated to match this.  Should I use a pris.ly url, or would should I just use the github URL instead?

hayes:
Cool, renamed the examples, and then refactored the pothos one a bit:

* Enabled testing
* Split out the schema into multiple typescript files
* Started printing out the graphql schema

typescript/graphql-pothos/src/server.ts
janpio: I think we always use the pris.ly URLs here @nikolasburk ?

typescript/graphql/src/schema/post.ts
```diff
@@ -0,0 +1,181 @@
...
+      id: t.arg.int({ required: true }),
+    },
+    resolve: (query, parent, args) => {
+      return prisma.post.update({
+        ...query,
+        where: { id: args.id },
+        data: { published: true },
+      })
+    },
+  }),
```
nikolasburk:
```suggestion
  togglePublishPost: t.prismaField({
    type: 'Post',
    args: {
      id: t.arg.int({ required: true }),
    },
    resolve: async (query, parent, args) => {
      // Toggling become simpler once this bug is resolved: https://github.com/prisma/prisma/issues/16715
      const postPublished = await prisma.post.findUnique({
        where: { id: args.id},
        select: { published: true }
      })
      console.log(postPublished)
      return prisma.post.update({
        ...query,
        where: { id: args.id },
        data: { published: !postPublished?.published },
      })
    },
  }),
```

typescript/graphql-pothos/src/server.ts
nikolasburk: I will take care of creating the new pris.ly link for Nexus and update the Nexus example accordingly 👍

nikolasburk:
Thanks so much for contributing this, Michael! ❤️ The `graphql` example is one of the most popular ones in the entire repo, so this is a really impactful change 🙌

PR #4363, opened 2023-01-02
fix(example): Add typing and remove unused imports

Touratica:
Adds typing to options and removes unused import of Providers from next-auth/providers


ruheni:
Thanks for the contribution 🙏🏽

Going ahead and merging this.

PR #4403, opened 2023-01-19
feat: Improve rest-nextjs-api-routes example

Touratica:


ruheni:
Hey 👋🏽

Thanks for looking into this and keeping the Next.js example up to date with the current standards. 🙂

As a small token of appreciation, could you message me (Twitter or Public Prisma Slack) your address and t-shirt size so that we can send you some swag?

PR #4557, opened 2023-03-06
Fix Scoping bug

warg-void:
Fixes the scoping bug in props under the javascript nextjs example, which is causing the entire page (http://localhost:3000/p/1) to not read the data correctly.

This is the bug (happens when you click on any blog post or draft):

image


ruheni:
Thank you for catching this and your contribution 💚

PR #4751, opened 2023-06-01
(Don't merge): try ESM prototype

Jolg42:
Related to https://github.com/prisma/prisma/pull/19184


Jolg42:
There is another PR that replaced the first one https://github.com/prisma/prisma/pull/21858

We can close this for now.

PR #5208, opened 2023-09-11
feat: add accelerate next.js starter project

ankur-arch:
- [x] Update root README to include PDP-related projects
- [x] Add Accelerate starter project using Next.js app router and edge functions


nikolasburk:
Thanks for this PR, Ankur! Let's rather turn the Accelerate example into its own repo similar to the Pulse examples. That way, we can also add a "Deploy with Vercel" button to the README.

janpio:
> That way, we can also add a "Deploy with Vercel" button to the README.

I have some vague recollection that Vercel added something to also deploy subfolders of repos now, so maybe that would also be possible here now. Double check though.

PR #5326, opened 2023-10-03
fix: Drafts is not set as the default export

skyddyyu:
The Drafts page does not have export set causing an error when routing to that page


ruheni (approved):

ruheni:
Thanks for catching this and resolving it!

PR #5782, opened 2024-03-21
add teable

ankur-arch:
Adding [Teable](https://github.com/teableio/teable) to Prisma real-world OSS.


nikolasburk (approved):
looks awesome!!

PR #5932, opened 2024-05-07
Update seed.ts

oasaleh:
No need to import this.

### DIFF
--- a/typescript/rest-nextjs-api-routes-auth/prisma/seed.ts
+++ b/typescript/rest-nextjs-api-routes-auth/prisma/seed.ts
@@ -1,4 +1,4 @@
-import { PrismaClient, Prisma } from '@prisma/client'
+import { PrismaClient } from '@prisma/client'
 
 const prisma = new PrismaClient()
 


typescript/rest-nextjs-api-routes-auth/prisma/seed.ts
```diff
@@ -1,4 +1,4 @@
-import { PrismaClient, Prisma } from '@prisma/client'
+import { PrismaClient } from '@prisma/client'
```
jharrell, line 1: Hi there, we do need that import for the type on line 5. 

```suggestion
import { PrismaClient } from '@prisma/client'
import type { Prisma } from '@prisma/client'
```

oasaleh, line 1: Hmm I didn't get an error by removing it. Also, it wasn't imported as "type" before. Are both imports equal? (the one with type and the one without?)

jharrell: The tests on this PR are failing, unfortunately, so we can't remove it. 

They're not strictly equivalent but it works similar/the same

jharrell:
Hi there! This is needed for the typing on line 5 https://github.com/prisma/prisma-examples/pull/5932/files#diff-092b09d9d76b70428733410975751e3db3ba47a581816ac3d3e95f34d80e420aR5

Thank you very much for the pull request, but we'll be closing for now.

PR #6020, opened 2024-05-29
add BigQuery example for Pulse

nikolasburk:


pulse/data-sync-with-bigquery/index.ts
```diff
@@ -0,0 +1,89 @@
...
+  for await (const event of stream) {
+    console.log('Just received an event:', event)
+    const user: User = event.created
+    syncData(user)
+  }
+}
+
+
+
+async function syncData(user: User) {
```
ankur-arch, line 39: I'd avoid storing the email from the user object. Due to compliance [issues](https://medium.com/@florian.trehaut/ensuring-gdpr-compliance-on-gcp-bigquery-efficiently-managing-the-right-to-be-forgotten-a76137944633) popping regarding storing PIIs such as email.

nikolasburk: I would say that's ok for the demo. The point of this example is to show the _minimal mechanics_ of syncing data into BigQuery with Pulse, I don't think anyone will use this exact example but people will rather use this as an inspiration and building block to sync whatever business data _they_ have.

PR #6066, opened 2024-06-10
enhance: improve guidance on example #DA-326

ankur-arch:


databases/postgresql-supabase/README.md
```diff
@@ -88,10 +88,16 @@ Your database is now in sync with your schema.
 
 ### 3. Run the script
 
-To run the script `script.ts`, run the following command: 
+For the script to work, you first need to execute the [seed](./prisma/seed.ts) script to seed your database. You can do that using the following command:
+
+```
+npx prisma db seed
+```
```
janpio, line 95: Is this always needed or does `migrate dev` maybe take care of that if you actually use it?

ankur-arch: AFAIK it should run with the migrate or generate script, but it did not. Hence I included it. 

image

ankur-arch: It says [here](https://www.prisma.io/docs/orm/prisma-migrate/workflows/seeding#integrated-seeding-with-prisma-migrate) it should run automatically given the `seed` property exists.

ankur-arch: When I run `npx prisma migrate reset` the seed script executes automatically.

image

janpio: Yeah, I recently had a conversation with @nikolasburk that the docs are not clear enough there. It is only run in `migrate dev` when that also runs a reset behind the scenes because of drift and you accepting that. It does _not_ generally run it.

So our instructions were definitely missing that until now.

ankur-arch: @janpio so does this look good to you? If so let me or @jharrell know so we can merge it.

janpio: The `doesUserExist` bit looks weird to me as a good portion, especially at the beginning, of our example code is how to avoid ending up in an error situation, but if that is how the example works, then that is how it has to be.

databases/postgresql-supabase/script.ts
```diff
@@ -3,6 +3,22 @@ import { PrismaClient } from '@prisma/client'
...
+  })
+
+  if(!doesUserExist) {
+    console.log('To run this example, please execute the seed script first.');
+    console.log('This script will add a user with the email alice@prisma.io to the database.');
+    console.log('The user is required for one of the queries.');
+    console.log('You can run the seed script using the following command:');
+    console.log('npx prisma db seed');
+    return
+  }
```
jharrell:
```suggestion
  try {
    await prisma.user.findFirstOrThrow({ where: { email: 'alice@prisma.io' } })
  } catch {
    const consoleMessage = `
User alice@prisma.io not found. Please run the seed script before running this script.
You can run the seed script via the following command:

npx prisma db seed
`;
    console.error(consoleMessage);
    return
  }
```

ankur-arch: I like your suggestion @jharrell 👍

databases/postgresql-supabase/README.md
ankur-arch: @jharrell resolved that.

jharrell (approved):

PR #6151, opened 2024-06-25
Added Remix Accelerate Example App

nurul3101:
This completes DA-880


accelerate/remix-starter/app/components/Quote/Quote.tsx
```diff
@@ -0,0 +1,80 @@
...
+  const [{ id, quote, createdAt }, { cacheStatus, region, lastModified }] = [
+    result.data,
+    result.info,
+  ];
+
+  return (
+    <QuoteWrapper title={title} type={type}>
+      <div className="flex flex-col">
+        <p className="text-lg">
+          {" "}
```
jharrell: I see we have this in some `p` tags and not others. Any reason we need this whitespace?

accelerate/remix-starter/app/data/query.server.ts
```diff
@@ -0,0 +1,25 @@
...
+      // You can find the `cacheStrategy` options [here](https://www.prisma.io/docs/accelerate/caching#cache-strategies). The `cacheStrategy` can also be undefined, which would mean only connection pooling is being used.
+      cacheStrategy: strategy,
+      orderBy: {
+        id: "desc",
+      },
+      take: 1,
+    })
+    .withAccelerateInfo();
+
+  console.log(result);
```
jharrell: Should probably remove console logs (unless this is important for the demo)

accelerate/remix-starter/app/data/utils/prisma.server.ts
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// This is needed because in development we don't want to restart
+// the server with every change, but we want to make sure we don't
+// create a new connection to the DB with every change either.
+if (SERVER_ENV.NODE_ENV === "production") {
+  prisma = buildClient();
+  prisma.$connect();
+} else {
+  if (!(global as any).__prisma) {
```
jharrell: I'd generally like to avoid `as any`. Is that possible to remove/what issues are we running into that requires this?

accelerate/remix-starter/app/data/query.server.ts
nurul3101: Thanks for noticing! Removed the log.

accelerate/remix-starter/app/components/Quote/Quote.tsx
nurul3101: I added whitespace through CSS. Good catch.

accelerate/remix-starter/app/data/utils/prisma.server.ts
nurul3101: I had taken reference from our accelerate/nextjs-starter project in which it was defined this way, so kept the same way. I have made the changes to remove `as any`.

accelerate/remix-starter/README.md
```diff
@@ -0,0 +1,99 @@
...
+
+# Accelerate connection string (used for queries by Prisma Client)
+DATABASE_URL="__YOUR_ACCELERATE_CONNECTION_STRING__"
+
+# Database connection string (used for migrations by Prisma Migrate)
+DIRECT_DATABASE_URL="__YOUR_DATABASE_CONNECTION_STRING__"
+
+```
+
+Note that `__YOUR_DATABASE_CONNECTION_STRING__` and `__YOUR_ACCELERATE_CONNECTION_STRING__` are placeholder values that you need to replace with the values of your database and Accelerate connection strings. Notice that the Accelerate connection string has the following structure: `prisma://accelerate.prisma-data.net/?api_key=__YOUR_ACCELERATE_API_KEY__`.
```
jharrell, line 45: Instructions need to be updated as this isn't what the `.env.example` looks like anymore.

jharrell (dismissed):

accelerate/remix-starter/README.md
nurul3101: Nice catch! I updated the .env.example so that Instructions are accurate.

jharrell (approved):

PR #6172, opened 2024-06-28
Feat solidstart starter

ludralph:
This completes DA-882


accelerate/solidstart-starter/README.md
```diff
@@ -0,0 +1,85 @@
...
+
+### 5. Start the app
+
+You can run the app with the following command:
+
+```
+npm run dev
+```
+
+You can see the performance and other stats (e.g. cache/hit) for the different Accelerate cache strategies at the bottom of the UI:
```
ankur-arch, line 91: There is no image 😕 . Plus I similar to the suggestion on the Svelkit and Nuxt example app to say what the app actually does.

ankur-arch (approved):

PR #6375, opened 2024-08-02
add typedsql example

jharrell:
Resolves DA-1250
Resolves DA-1011
Resolves DA-1012
Resolves DA-1013


nikolasburk (dismissed):
This is great, Jon 🎉  I ran the example and it worked flawlessly 👌 

I have a few notes:
- can we add a README file with the precise instructions to run this example?
- I'm against creating custom scripts for Prisma CLI commands like you did with `dbpush` and `generate`. in general with the projects in `prisma-examples`, I'm pushing for minimal layers of indirection/magic to teach people the basic workflows and let them decide what kind of abstractions/workflows they want to introduce on top.
- even if we import the typed query from `.prisma`, should we keep the `PrismaClient` import at `@prisma/client` as usual?
- to be consistent with the other examples, I think we should call the "run script" `dev` instead of `test` (or otherwise `start`, but `test` to me has the notion of executing actual _(unit or integration) tests_ which isn't really what this does).

typescript/typedsql/index.ts
```diff
@@ -0,0 +1,12 @@
+import { PrismaClient } from './node_modules/.prisma/client'
+import { conversionByVariant } from './node_modules/.prisma/client/sql'
```
nikolasburk: I don't think the `node_modules` prefix is needed.

```suggestion
import { PrismaClient } from `@prisma/client'
import { conversionByVariant } from '@prisma/client/sql'
```

nikolasburk:
ah never mind my comments about the `import`s, it seems like they will change to `@prisma/client/sql` anyways 🙏 

image

ankur-arch:
This works for me, well done @jharrell 👌 . 

Thinking similar to Niko here, a README for this similar to the starter script would be [helpful](https://github.com/prisma/prisma-examples/tree/latest/typescript/script). Where we guide users how to get started and show where the raw queries are.

.github/tests/typescript/typedsql/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eu
+
+npm install
+npm run dbpush
+npm run generate
+npm run test
```
jharrell:
```suggestion
npm run dev
```

.github/tests/typescript/typedsql/run.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+
+set -eu
+
+npm install
+npm run dbpush
+npm run generate
+npm run dev
```
jharrell:
```suggestion
npm install
npx prisma db push --skip-generate
npx prisma generate --sql
npm run dev
```

mhessdev (commented):
Should we add sqlite to the folder name? then we could make others for postgres and MySQL.

mhessdev:
> Should we add sqlite to the folder name? then we could make others for postgres and MySQL.

@ankur-arch lmk what you think as well

mhessdev (dismissed):

ankur-arch (dismissed):

mhessdev (approved):

PR #6526, opened 2024-08-24
Add simple accelerate starter

petradonka:


jharrell:
@sneub since you implemented the typescript and pulse starter, could you give this a review as well? Your perspective would be nice 😊

jharrell (approved):

PR #6979, opened 2024-11-01
roll back versions in trpc-nextjs

jharrell:


nikolasburk (dismissed):
I tested the app and it all worked with the new versions 🎉 

Small note: Maybe also fix this typo with the two extra asterisks in the README: 


image

nikolasburk (approved):

PR #6990, opened 2024-11-04
Add Render example to Deployment platforms

jess-render:
## About

This PR adds a Render deployment example to the prisma-examples repo.
This example pairs with the docs in https://github.com/prisma/docs/pull/6396.


deployment-platforms/render/package.json
```diff
@@ -0,0 +1,20 @@
...
+    "dev": "PORT=3000 node src/index.js",
+    "start": "node src/index.js"
+  },
+  "author": "@jess-render",
+  "license": "ISC",
+  "dependencies": {
+    "@prisma/client": "5.21.1",
+    "body-parser": "1.20.3",
+    "express": "4.21.1",
+    "prisma": "5.21.1"
```
jharrell, line 20: prisma is usually a dev dependency

deployment-platforms/render/src/index.js
```diff
@@ -0,0 +1,149 @@
...
+const app = express()
+
+app.use(bodyParser.json())
+app.use(express.static('public'))
+
+app.get(`/api`, async (req, res) => {
+  res.json({ up: true })
+})
+
+app.get(`/api/seed`, async (req, res) => {
```
jharrell: interesting! We also have seed functionality built into Prisma ORM: https://www.prisma.io/docs/orm/prisma-migrate/workflows/seeding

Did that not work in this case?

jharrell (dismissed):
LGTM, just had a few clarifying questions 😊

deployment-platforms/render/package.json
jess-render: I did this for a very specific reason... but with a bit of digging, I found a better solution. Will update!

Long version:
Today Render doesn't have a distinction between "build dependencies" and "prod dependencies".

If someone sets NODE_ENV to `production` in their service, npm install would skip the devDependencies, which are needed for the build step.

But TIL about `--production=false`

> With the --production flag (or when the NODE_ENV environment variable is set to production), npm will not install modules listed in devDependencies. To install all modules listed in both dependencies and devDependencies when NODE_ENV environment variable is set to production, you can use --production=false.
from https://docs.npmjs.com/cli/v10/commands/npm-install

deployment-platforms/render/src/index.js
jess-render: Thanks, I'll update this.

I didn't know about the seed feature. (I actually copied the existing Heroku example to keep these two examples easier to compare.)

jharrell: Ah, that makes sense. I think the heroku example is a bit dated. Understand why you went this path though 🙏

jess-render: Just updated the code with a seed script. Also added a "Seed the database" section to the docs side. :)

jharrell (approved):

PR #7019, opened 2024-11-08
added option to use PPG to READMEs

nikolasburk:


.github/readmes/orm/_setup-3.md
```diff
@@ -3,12 +3,43 @@ npm install
...
+### [Optional] Switch database to Prisma Postgres
+
+This example uses a local SQLite database by default. If you want to use to [Prisma Postgres](https://prisma.io/postgres), follow these instructions (otherwise, skip to the next step):
+
+1. Set up a new Prisma Postgres instance in the Prisma Data Platform [Console](https://console.prisma.io) and copy the database connection URL.
+2. Update the `datasource` block to use `postgresql` as the `provider` and paste the database connection URL as the value for `url`:
+    ```prisma
+    datasource db {
+      provider = "postgresql"
+      url      = "prisma+postgres://accelerate.prisma-data.net/?api_key=ey...."
```
jharrell, line 15: Should we advise this instead of a .env file?

jharrell (dismissed):

.github/readmes/orm/_setup-3.md
nikolasburk: I may be alone with this opinion but I'm against that. The main reason being _simplicity_ and _reducing layers of indirection_ which to me has always been one of the most important things about the examples. 

A few more thoughts:
- the examples should be minimal, allowing people to understand how Prisma fits together with a certain framework/technology without needing to look at many files (that's why most examples have all their code in a single file)
- the tradeoff is "simplicity" vs "best practices"; personally I'm willing to accept it
- another thesis with the minimalism in this repo is that users have their own preferences for how to structure their apps; e.g. they may not want to use `.env` for setting their env vars 

All of that is definitely debatable, just sharing my personal views and willing to change if more people think we should use `.env` 🙂 

So, I personally would not want to "clutter" the instructions for setting up PPG with an instruction to create an `.env` file. 

As a compromise, we could add a sentence that in production it's recommended to not hardcode connection strings but rather use env vars, e.g. via `.env` files.

ankur-arch (approved):

PR #7731, opened 2025-03-24
feat: DA-3001 Solid Start Example Project

aidankmcalister:


orm/solid-start/lib/prisma.ts
```diff
@@ -0,0 +1,6 @@
+import { PrismaClient } from '@prisma/client'
+import { withAccelerate } from '@prisma/extension-accelerate'
+
+const prisma = new PrismaClient().$extends(withAccelerate()) // If you're not using Prisma Postgres, remove `.$extends(withAccelerate())`
```
jharrell: We can remove the comment as it's in our readme.
```suggestion
const prisma = new PrismaClient().$extends(withAccelerate())
```

orm/solid-start/README.md
```diff
@@ -0,0 +1,160 @@
...
+
+### 3. Start the SolidStart server
+
+```
+npm run dev
+```
+
+The server is now running at http://localhost:3000. You can now view your todo list!
+
+## Switch to another database (e.g. PostgreSQL, MySQL, SQL Server, MongoDB)
```
jharrell: I see why you did this and we may have done it with other examples, but I think a better solution here would be to just reference our documentation on how someone can use a different db engine/provider.

orm/solid-start/prisma/schema.prisma
```diff
@@ -0,0 +1,30 @@
+// This is your Prisma schema file,
+// learn more about it in the docs: https://pris.ly/d/prisma-schema
+
+// Looking for ways to speed up your queries, or scale easily with your serverless or edge functions?
+// Try Prisma Accelerate: https://pris.ly/cli/accelerate-init
+
+generator client {
+  provider = "prisma-client-js"
```
jharrell, line 8: let's define a custom output path as that will become the default soon. (make sure you also add it to `.gitignore`.

jharrell (approved):

PR #7732, opened 2025-03-26
all examples with explicit output

jharrell:
Resolves DA-2811

Repos that I couldn't update due to incompatibilities with vite:
- nuxt
- react-router-7
- sveltekit

Repos I couldn't update, unclear why:
- nest
- nest-graphql
- nest-graphql-sdl-first

We should try updating this post 6.6.0 w/ the new `prisma-client-ts` generator: https://linear.app/prisma-company/issue/DA-3034/update-remaining-prisma-example-repos-with-new-generator


jkomyno:
Note: the generator is now called `prisma-client`.

Issue #8280, opened 2025-08-19
Nuxt, generator-prisma-client: Invalid `prisma.quotes.findMany()` invocation in production build


tobiasdiez:
Reproduce:
```
cd generator-prisma-client/nuxt3-starter-nodejs
pnpm install
pnpm build
pnpm preview
```

yields:
```
PrismaClientKnownRequestError:
Invalid `prisma.quotes.findMany()` invocation:


Cannot find module 'D:\Programming\prisma-examples\generator-prisma-client\nuxt3-starter-nodejs\.output\server\node_modules\@prisma\client\runtime\query_compiler_bg.postgresql.wasm'
    at Yn.handleRequestError (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/node_modules/@prisma/client/runtime/client.mjs:88:7459)
    ... 8 lines matching cause stack trace ...
    at async C (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/chunks/_/nitro.mjs:672:7219) {
  cause: PrismaClientKnownRequestError:
  Invalid `prisma.quotes.findMany()` invocation:


  Cannot find module 'D:\Programming\prisma-examples\generator-prisma-client\nuxt3-starter-nodejs\.output\server\node_modules\@prisma\client\runtime\query_compiler_bg.postgresql.wasm'
      at Yn.handleRequestError (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/node_modules/@prisma/client/runtime/client.mjs:88:7459)
      at Yn.handleAndLogRequestError (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/node_modules/@prisma/client/runtime/client.mjs:88:6784)
      at Yn.request (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/node_modules/@prisma/client/runtime/client.mjs:88:6491)
      at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
      at async l (file:///D:/Programming/prisma-examples/generator-prisma-client/nuxt3-starter-nodejs/.output/server/node_modules/@prisma/client/runtime/client.mjs:97:9772)
[...]

PR #8281, opened 2025-08-19
Nuxt, generator-prisma-client: Use experimental wasm mode of nitro

tobiasdiez:
`pnpm build` works but preview results in the same issue as https://github.com/prisma/prisma-examples/issues/8280. So it's likely that the wasm file is actually never processed by nitro.


AmanVarshney01:
Hey, thanks @tobiasdiez. This should be fixed now. Feel free to reopen if you’re still having problems.

PR #8368, opened 2025-11-08
Update Nuxt & React Router examples

AmanVarshney01:


generator-prisma-client/nuxt3-starter-nodejs/prisma/schema.prisma
```diff
@@ -1,13 +1,12 @@
...
+  output     = "../lib/generated/prisma"
+  engineType = "client"
 }
 
 datasource db {
-  provider  = "postgresql"
-  url       = env("DATABASE_URL")
-  directUrl = env("DIRECT_URL")
+  provider = "postgresql"
+  url      = env("DATABASE_URL")
```
mhartington, line 9: We can remove this from the schema now

mhartington (commented):
We can remove the `url` from the prisma schema as per the latest pre-release. Other than that, it looks good!

PR #8372, opened 2025-11-10
Update misc examples

AmanVarshney01:


mhartington (commented):
We can remove the url from the prisma schema as per the latest pre-release.

PR #8564, opened 2026-07-28
fix(compute): add Hono template entrypoint

AmanVarshney01:
## Purpose of change

Make the Hono Prisma Compute template deployable by giving the Bun build strategy an explicit source entrypoint.

## What's changed and why

- Add `entry: "src/index.ts"` to `compute/hono/prisma.compute.json`. Without it, the template installs, applies its database schema, and compiles successfully, but artifact creation fails with `Entrypoint is required` because `package.json` has no `main` or `module` field.

## Testing notes / Before-after

- Before: `app build` failed after TypeScript compilation with `Entrypoint is required. Pass --entrypoint or define package.json main`.
- After: `app build --json --no-interactive` succeeds with build type `bun` and packaged entrypoint `index.js`.
- `npx vitest run tests/compute.test.ts`
- `npx tsc --noEmit`
- `DATABASE_URL=postgresql://user:pass@localhost:5432/db bun run build`

### DIFF
--- a/compute/hono/prisma.compute.json
+++ b/compute/hono/prisma.compute.json
@@ -2,6 +2,7 @@
   "app": {
     "name": "prisma-compute-hono",
     "framework": "hono",
+    "entry": "src/index.ts",
     "httpPort": 8080
   }
 }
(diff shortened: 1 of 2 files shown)


AmanVarshney01:
@coderabbitai approve

PR #8596, opened 2026-08-26
fix(compute): add baseline migrations to database templates

AmanVarshney01:
## Problem

Composer 0.15 provisions the database and Compute service, then replays the migration path committed with the application. The Hono, Next.js, and TanStack Start templates contained a contract and generated contract artifacts, but no authored path from an empty database to that contract. A first deployment therefore failed during the migration resource with `MIGRATION_PATH_NOT_FOUND`.

The templates also still pinned `@prisma/orm-postgres` 8.0.0-rc.4, while `@prisma/composer-prisma-cloud` 0.15.0 requires 8.0.0-rc.8. That version alignment changes the emitted contract, including the temporal codec IDs, so the contracts and migration baselines need to be generated together with RC.8.

## Fix

- align `@prisma/orm-postgres` and the Bun lockfiles on 8.0.0-rc.8
- regenerate the committed contract artifacts
- add an initial migration and target snapshot for Hono, Next.js, and TanStack Start, authored with `prisma contract emit` and `prisma migration plan --name init`
- require database templates to include migrations and run Prisma's offline migration graph integrity check in CI

This uses the normal Prisma migration workflow; it does not add deploy-time schema generation or a Composer workaround.

## Verification

- `npx vitest run tests/compute.test.ts`
  - frozen installs with Bun 1.3.14
  - builds all Compute templates
  - verifies generated contracts remain clean
  - runs `prisma migration check --space app` for every database template
  - assembles and boots every Composer service bundle

All checks pass locally.

## Summary by CodeRabbit

* **New Features**
  * Added initial database migrations for users and posts, including relationships, uniqueness rules, indexes, and default values.
  * Added migration tracking and schema validation across supported application templates.
* **Bug Fixes**
  * Improved PostgreSQL date and timestamp handling with more precise temporal representations.
  * Updated automatic timestamp generation for created and updated records.
* **Chores**
  * Updated the PostgreSQL ORM integration to the latest release candidate.


AmanVarshney01:
@coderabbitai approve

